PR #4, opened 2018-08-09
Update talks

wbhart:


wbhart:
Unfortunately squashing all my commits means they will forever show up when I push anything to my repo.

Obviously only the last commit is relevant.

PR #6, opened 2018-08-17
Absolute URLs

ederc:
I think this should fix this absolute url problems. If I forgot some, please point them out to me and I will fix them, too.


sebasguts (approved):
Nice, thank you!

PR #47, opened 2019-06-07
Added talk santiago

sebasguts:


wbhart:
I'm pleased this particular speaker is giving a talk, but gosh we really need to start having meetings from Monday morning to Friday afternoon if we want to load them up with talks and still expect to get work done. There's now a proposal for two additional people to give talks!

PR #68, opened 2020-01-30
install.md: reduce number of .deb packages to install

fingolfin:
Most of our packages are now shipping with BinaryBuilder enabled, so those dependencies are not all necessary anymore. The notable exception is GAP.jl, which only needs the limited list I left in.

Of course one might be worried about people who are following these instructions on exotic architectures were we don't make any binary builders available. But for those also the instructions to install Julia are broken...

Also, perhaps I am missing something, and we do need some of those after all... But what then am I missing?

### DIFF
--- a/install.md
+++ b/install.md
@@ -55,7 +55,7 @@ Now update and install the necessary system packages.
 
 {% highlight bash %}
 apt-get update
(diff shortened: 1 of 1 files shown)


frankluebeck:
The gap repo also mentions
```
zlib1g-dev
```
as a dependency, and a GAP user has recently figured out that
```
libboost-all-dev libczmq-dev libcurl4-gnutls-dev libcdd-dev
```
are also needed to build all current GAP packages.

fingolfin:
Zlib is optional for compiling GAP, as we bundle it.

Compiling all GAP packages is IMHO neither necessary nor desirable. Also, boost is not needed anymore, but in your package list it presumably pulls in a bunch of dependencies that in turn are needed by GAP packages.

Issue #72, opened 2020-01-31
Add all the missing talks from 2019 and 2020 -- and possibly disable the "Talks" page until we have any


fingolfin:
We currently only listed a single talk for 2019, which looks really bad. Luckily, it's also wrong, there were talks.

I believe it is the responsibility of the PIs and the people employed by the SFB to fix this, by (a) adding their talks, and (b) encouraging other people they know who talked about OSCAR to add theirs, too.

To do so, you can edit the file https://github.com/oscar-system/oscar-website/blob/gh-pages/_data/talks.yml which is fairly straight forward. If anything is unclear, though, don't hesitate to ask.

In the meantime, unless we can very quickly populate this, I think we should disable it for the time being. Better to list no talks than this outdated list. See PR #73

fingolfin:
In addition, we need to designate somebody to make sure we don't fall back so badly again. In the past, Sebastian did that...

fingolfin:
@fieker You gave a talk on OSCAR at CIRM in January 2019, right? Could you please make sure it is added here?

@micjoswig You gave a presentation on OSCAR just recently at the MPI. Could you please make sure it is added here?

PR #73, opened 2020-01-31
Disable talks page for now

fingolfin:
See also issue #72


fingolfin:
We have some content now

PR #75, opened 2020-02-04
Fix a bunch of links and URLs

fingolfin:
There were lots of 404 errors for the homepages of project members; seems TU Kaiserslautern shuffles URLs randomly every couple months? I fixed the links to @fieker and @wdecker but could not find any homepages for @wbhart @thofma @rbehrends @CarloSircana @hannes14 so I just disabled those for now.  We could instead point at their GitHub pages, or the TU Kaiserslautern could resume providing homepages for all researchers they employ *shrug*, that's not something I can influence (yet!).

Also, I think Andreas Steenpaß isn't at the TU Kaiserslautern anymore; perhaps we should point that out?

Besides that, I changed a bunch of URLs from http to https (unfortunately, this is not supported by the homepages of Arb, Flint, Nemo, and a few others).

Finally, I fixed the homepage URL in the `_config.yml`, it was still pointing at @sebasguts homepage (shouldn't really matter for deployment, though).

### DIFF
--- a/_data/contributors.yml
+++ b/_data/contributors.yml
@@ -4,7 +4,7 @@
 
 - name: Reimer Behrends
   affiliation: TU Kaiserslautern
-  website: http://www.mathematik.uni-kl.de/agag/mitglieder/wissenschaftliche-mitarbeiter/dr-reimer-behrends/
+  #website: https://www.mathematik.uni-kl.de/agag/mitglieder/wissenschaftliche-mitarbeiter/dr-reimer-behrends/
   paid_by_dfg: true
 
 - name: Alex Best
@@ -19,11 +19,11 @@
 
 - name: Wolfram Decker
   affiliation: TU Kaiserslautern
-  website: http://www.mathematik.uni-kl.de/en/agag/members/professors/prof-dr-wolfram-decker/
+  website: https://www.mathematik.uni-kl.de/en/agag/people/head/prof-dr-wolfram-decker/
 
 - name: Christian Eder
   affiliation: TU Kaiserslautern
-  website: http://www.mathematik.uni-kl.de/~ederc/index.html
+  website: https://www.mathematik.uni-kl.de/~ederc/index.html
 
 - name: Raul Epure
   affiliation: TU Kaiserslautern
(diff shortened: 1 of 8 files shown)


wbhart:
TUK does provide websites for us. But they don't know anything about SEO (amongst other things). For more than 6 months, Google hasn't been able to find them. Cuts on spam anyway.

It's a little more complicated than this, actually. But that's better for an in person conversation.

wbhart:
Indeed Steenpas is not longer at TUK.

_data/contributors.yml
```diff
@@ -50,7 +50,7 @@
   
 - name: Tommy Hofmann
   affiliation: TU Kaiserslautern
-  website: http://www.mathematik.uni-kl.de/agag/mitglieder/wissenschaftliche-mitarbeiter/thofmann/
```
thofma, line 53: https://www.mathematik.uni-kl.de/~thofmann/

wbhart:
Let me clarify. Google now finds the "homepage" for us, but it is not really a homepage. This is intentional and was voted on. Details can be discussed in person.

I'm not suggesting I agree with it of course.

Of course you can always link to my page here :-)

http://homepages.warwick.ac.uk/~masfaw/

fieker:
On Tue, Feb 04, 2020 at 05:05:03AM -0800, wbhart wrote:
> Let me clarify. Google now finds the "homepage" for us, but it is not really a homepage. This is intentional and was voted on. Details can be discussed in person.
> 
> I'm not suggesting I agree with it of course.
> 
> Of course you can always link to my page here :-)
> 
> http://homepages.warwick.ac.uk/~masfaw/

It about as up-to-date as the rst of the home pages...

wbhart:
To be completely fair about it, the GDPR and a new CMS have made this difficult enough to put basically everyone off doing anything about it.

_data/contributors.yml
fingolfin: @thofma changed

fingolfin:
Honestly, I don't really care about whether people in Kaiserslautern have a homepage or not or where, I just don't want broken links on the website; hence this PR. Can we just merge it, and then if people want a link to their homepage / want a different link / want no link, they can do so via pull requests or commits of their own?

Of course I could have pushed this myself directly without telling anybody, but I thought it would be good to run it by the affected people...

wbhart:
Fine by me.

PR #79, opened 2020-02-04
Don't duplicate talk dates

fingolfin:
This fixes one talk w/o date_sortable, and one where date_sortable and
date did not match.

### DIFF
--- a/_data/talks.yml
+++ b/_data/talks.yml
@@ -1,7 +1,6 @@
 - title: "Chevalley’s Theorem on constructible images made constructive"
   author: "Mohamed Barakat"
-  date: "January 30, 2020"
-  date_sortable: 20200130
+  date: "2020-01-30"
   conference: Computational and algorithmic methods
   conference_url: https://www.newton.ac.uk/event/graw02
   location: "Isaac Newton Institute, Cambridge, UK"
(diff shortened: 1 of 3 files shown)


_data/talks.yml
```diff
@@ -10,125 +9,109 @@
 
 - title: "OSCAR"
   author: "The OSCAR Team"
-  date: "February 18, 2019"
-  date_sortable: 20180925
+  date: "2019-02-18"
```
fingolfin, line 12: date_sortable above was wrong

_data/talks.yml
```diff
@@ -10,125 +9,109 @@
...
+  date: "2017-07-28"
   location: Kaiserslautern
 
 - title: "Basic arithmetic in Flint and Nemo"
   author: "William Hart"
-  date: "October 25, 2017"
+  date: "2017-10-23"
   location: Lyon
-  pdf: "RIAM-talk.pdf"
+  pdf: "RAIM-talk.pdf"
```
fingolfin, line 72: This talk had no `date_sortable` and thus did not appear on the talks page, which is probably why nobody noticed the broken link in it. The new `date` is based on what the PDF says.

PR #85, opened 2020-04-09
Remove broken GAP-Jupyter examples

### LINKED ISSUE #63: "Call Singular from GAP" notebooks do not work (anymore)
The code in the notebooks

https://nbviewer.jupyter.org/github/oscar-system/OSCARBinder/blob/master/GAPJulia_Singular_car.ipynb

and

https://nbviewer.jupyter.org/github/oscar-system/OSCARBinder/blob/master/GAPJulia_Singular_blog.ipynb

provided by @ThomasBreuer does not work anymore with recent versions.

fingolfin:
Resolves #63

Sad, but I see no alternative right now.


fieker:
If this works, I am fine with it. I feel that this is close to over-engineering. We replace a cron-job (in Siegen, but that is a different problem) with php scripts triggering systemd events. Is this maintainable?
In this form, at least, not by me. Since git does this all the time, I expect it is secure.

fingolfin:
@ThomasBreuer @fieker so, what about this PR? Do you agree to remove these examples?

fieker:
On Tue, Apr 21, 2020 at 04:17:02AM -0700, Max Horn wrote:
> @ThomasBreuer @fieker so, what about this PR? Do you agree to remove these examples?

Apart from me not being able to see anyhting in the pull request, be my
guest. If it is broken - and cannot be resurrected on short notice, its
history

mohamed-barakat:
The PR simply removes the two broken files.

PR #94, opened 2020-04-22
fix website for the "Factory" software

### LINKED ISSUE #91: Broken link for the Factory software
At https://oscar.computeralgebra.de/credits/, the link for Factory is ftp://www.mathematik.uni-kl.de/pub/Math/Singular/Factory, which currently doesn't work (since at least a few days).
I'm not sure who can fix that, so I will tentatively ask @hannes14 ?

rfourquet:
Closes #91.

### DIFF
--- a/_data/used_software.yml
+++ b/_data/used_software.yml
@@ -38,11 +38,10 @@
   website: http://home.imf.au.dk/jensen/software/gfan/gfan.html
 
 - name: Factory
-  website: ftp://www.mathematik.uni-kl.de/pub/Math/Singular/Factory
+  website: https://www.mathematik.uni-kl.de/ftp/pub/Math/Singular/Factory
 
 - name: cdd
   website: http://www-oldurls.inf.ethz.ch/personal/fukudak/cdd_home/
 
 - name: AbstractAlgebra
   website: https://github.com/Nemocas/AbstractAlgebra.jl
-


_data/used_software.yml
```diff
@@ -38,11 +38,10 @@
   website: http://home.imf.au.dk/jensen/software/gfan/gfan.html
 
 - name: Factory
-  website: ftp://www.mathematik.uni-kl.de/pub/Math/Singular/Factory
+  website: https://www.mathematik.uni-kl.de/ftp/pub/Math/Singular/Factory
```
fingolfin, line 41: @hannes14 is the link OK?

In any case, the link on https://www.singular.uni-kl.de/Manual/latest/sing_1.htm is also outdated (it's the same as the old one above)

fingolfin: I guess it could also link to https://www.singular.uni-kl.de/dox/html/factory_page.html

fingolfin: @hannes14 while we are at it, I made https://github.com/Singular/Sources/pull/993 (which does *not* attempt to resolve this URL, but a bunch of others on the same manual page)

hannes14 (approved):
the link is correct.
One may also add the link to the dox page, it contains the same README.
Difference: the first links has the tar balls, the docx link is part of Singulars doxygen pages.

rfourquet:
I will keep only this link then, I don't think we can provide two links for one software in this page.

PR #104, opened 2020-04-28
update CxxWrap workaround on MacOS for version 10

rfourquet:
In CxxWrap version 10, an override file has to be used instead of an environment variable.
cc @mohamed-barakat @benlorenz : does it look correct?
Preview at https://rfourquet.github.io/oscar-website/install/

### DIFF
--- a/install.md
+++ b/install.md
@@ -103,17 +103,22 @@ with Xcode 11.4 instead of using the binaries (replace <code>$SOMEWHERE</code> b
 the directory in which you wish to install <code>libcxxwrap-julia</code>,
 for example <code>$HOME/.julia/cxxwrap-workaround/</code>):
 {% highlight bash %}
-git clone -b v0.6.6 --depth 1 https://github.com/JuliaInterop/libcxxwrap-julia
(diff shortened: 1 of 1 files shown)


benlorenz:
Looks correct, yes.

mohamed-barakat:
Yes, thanks.

mohamed-barakat (approved):

rfourquet:
Ok thank you both!

fingolfin:
OTOH, https://github.com/oscar-system/Polymake.jl/pull/270 was merged -- so if there was another Polymake.jl release with it, we could just remove this workaround, no?

benlorenz:
> OTOH, [oscar-system/Polymake.jl#270](https://github.com/oscar-system/Polymake.jl/pull/270) was merged -- so if there was another Polymake.jl release with it, we could just remove this workaround, no?

Singular.jl needs the a similar workaround.

fingolfin:
Ah good point! Perhaps @hannes14 can look into it?

fingolfin (approved):

PR #105, opened 2020-04-29
add horizontal scrolling for code when it's too wide

### LINKED ISSUE #97: Code blocks are too narrow
And linebreaks are confusing. E.g. at https://rfourquet.github.io/oscar-website/install/ , section "Ubuntu 16.04 "Xenial": 

codeblocks

rfourquet:
The alternative consisting in letting long lines overflow their "parent" (I don't know the official word) is too ugly (achieved by removing `overflow: auto` from this PR):
scroll

And I didn't find how to automatically widen said "parent" to a big enough width to accomodate the width of code (and then, what to do when this would span wider than the screen itself?).

Fix #97.
As usual, preview at https://rfourquet.github.io/oscar-website/install/

### DIFF
--- a/public/css/poole.css
+++ b/public/css/poole.css
@@ -186,6 +186,10 @@ pre code {
 }
 .highlight pre {
   margin-bottom: 0;
+
+  /* add scrolling for code blocks which are too wide */
+  white-space: pre;
+  overflow: auto;
 }
 
 /* Gist via GitHub Pages */


fingolfin (approved):
Thanks! Works great in Firefox and Chrome. Unfortunately, not so in Safari. Dunno why, but I'd say that's no reason to withhold the fix from FF and Chrome uses/

fingolfin (commented):
Ah, I think it works in Safari if you remove the `word-wrap: break-word;` rule elsewhere, resp. override it here via `word-wrap: unset;`

public/css/poole.css
```diff
@@ -186,6 +186,10 @@ pre code {
 }
 .highlight pre {
   margin-bottom: 0;
+
+  /* add scrolling for code blocks which are too wide */
+  white-space: pre;
+  overflow: auto;
```
fingolfin, line 192: Note that in `poole.css`, we find in lines 169 and following:
```
  white-space: pre;
  white-space: pre-wrap;
  word-break: break-all;
  word-wrap: break-word;
...
```
which makes no sense... perhaps remove the `white-space: pre-wrap;` there and the `word-wrap`, and then add in the `overflow: auto` there?

rfourquet:
Ouch, I'm very sorry I pushed (from the command line) here instead of to my copy, and this merged it :-/
I had updated according to your suggestion.... Do you confirm that this new version works on Safari?

fingolfin:
It does, thanks!

Issue #98, opened 2020-04-24
Provide instructions how to take Julia-Oscar Jupyter notebooks and run them locally


fingolfin:
We need to improve the example page by giving instructions somewhere how one can download these notebooks and get them to run locally. That includes installing `Oscar.jl` (we already describe this), but also `IJulia`. Finally, how shall the user "load and run" that notebook.

Note that notebooks may have further requirements beyond Oscar and IJulia. Perhaps we can or should change `https://github.com/oscar-system/OSCARBinder` to have a `Project.toml` (or perhaps multiple, one for example: move the examples into a subdirs, with their own Project.toml. Perhaps even with a `Manifest.toml`; I don't know, figuring this out is part of the job).

@rfourquet we (the OSCAR PIs) thought that maybe you could work on that, given that you are  Julia expert. I don't know if you know anything about IJulia; but in a sense, if you don't, that might make you even more perfect for this... ;-)

fingolfin:
@rfourquet in the PI meeting, there was also an agreement that it is better to point people at the GitHub rendering of notebook rather than to Binder. But upon reviewing the links on the examples page, I am not 100% sure I understand our reasoning on this anymore: the motivation was that we prefer "dead" prerecorded notebooks over live ones, at least for now, because the experience with the "live" ones was that those have problems all the time (takes very long to start processing; has issues because the docker image is outdated; doesn't work as it worked on our own computer; etc.).

While I understand these concerns and agree with them, it seems to me that the current links to https://nbviewer.jupyter.org are (a) pointing to "non-live" versions of the notebooks anyway, and (b) are snappy. So maybe I am missing something? @micjoswig ?


(Of course at some point in the future, we'd want something live, at least *I* would want that, but the technology better be rock solid for that, and performant, and right now this simply is not a priority and we have to direct our limited resources carefully)

rfourquet:
The "dead" version (what is linked to currently) is indeed very snappy and seems appropriate, more so for "general" users than the rendered version on github (which is also great, but by default users won't be interested in github, and there is already a github link for those interested).

And I agree that the live version is quite slow to launch; that said, before it fully comes live, it's possible to start reading the notebook. But currently, I get a "kernel not found error"... so yes this seems premature for having the live version as default.

aaruni96:
We already have this text on the website :
![image](https://github.com/user-attachments/assets/b447d57d-043b-4833-827f-7fbf2471fc85)

[...]

PR #107, opened 2020-05-03
add instructions to run notebooks locally

rfourquet:
Partly addresses #98 (this does nothing about adding Project.toml files etc.)
I'm not sure if how I put it is not too much hand holding...

Preview at https://rfourquet.github.io/oscar-website/example/. These new instructions might look too prominent compared to the notebook link themselves.

### DIFF
--- a/example.html
+++ b/example.html
@@ -6,7 +6,8 @@ title: OSCAR examples
 Clicking on the example will open a static version of the example, powered by <a href="https://nbviewer.jupyter.org/">nbviewer</a>.
 To interact with a "live" version, it is possible from there to download the notebook and run it locally:
 <ul>
-  <li>right-click on the download icon of the notebook (in the upper right corner), which will allow
+  <li>open the context menu (via right-click on Linux and Windows, or ctrl-click on macOS)
+    on the download icon of the notebook (in the upper right corner), which will allow
     you to save the notebook file (say <code>Singular.ipynb</code>) in a directory of your choice
     (say <code>~/Documents</code>);</li>
   <li>make sure that you have the <code>Oscar</code> package installed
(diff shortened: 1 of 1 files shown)


example.html
```diff
@@ -4,6 +4,28 @@
 ---
 
 Clicking on the example will open a static version of the example, powered by <a href="https://nbviewer.jupyter.org/">nbviewer</a>.
+To interact with a "live" version, it is possible from there to download the notebook and run it locally:
+<ul>
+  <li>right-click on the download icon of the notebook (in the upper right corner), which will allow
```
rfourquet: What kind of click is it on Mac?

fingolfin: right click if one has a mouse or knows how to do this on the touchpad; otherwise ctrl-click ; might also say something like:
> open the context menu on the download (via right-click on Linux and Windows, or ctrl-click on macOS)

example.html
```diff
@@ -4,6 +4,28 @@
 ---
 
 Clicking on the example will open a static version of the example, powered by <a href="https://nbviewer.jupyter.org/">nbviewer</a>.
+To interact with a "live" version, it is possible from there to download the notebook and run it locally:
```
fingolfin, line 7: BTW how about converting this file from HTML to markdown (one can embed HTML into a `md` file, so the `<div class="gallery_example">` stuff at the end could probably stay unchanged).

At least for me, editing Markdown feels more convenient. But that's taste, so don't feel obliged to do this.

example.html
rfourquet: Ok thanks

example.html
rfourquet: Yes absolutely! I feel quite uncomfortable writing HTML when Markdown would seem to suffice. I will investigate how to do that (but not in this PR).

rfourquet:
I will try to find a way to make these instructions less prominent (e.g. hidden by default with just a pointer would be nice).

PR #114, opened 2020-05-13
Added GIT-Fans example.

rbehrends:


rfourquet:
In the rendered version at https://nbviewer.jupyter.org/github/oscar-system/GITFans/blob/master/ipynb/GITFans.ipynb, in `Out[12]`, an image seems to be missing. Is this easily fixable?

rbehrends:
@ThomasBreuer @jankoboehm Any ideas what caused that? It's also not visible in the GitHub rendering (and I have to admit that I didn't notice it myself).

ThomasBreuer:
I do not understand this.
When I run the cells in the notebook, I can see the graph in the cell of `Out[12]`.
Then I save the notebook, and `diff` shows that the coordinates of the points of the graph are now different, that's all.
Then I start `jupyter notebook` anew, load the notebook, and the graph is visible.
Is it possible that I can see the graph because of some cached data (in the browser?), which are missing at oscar-system?

rbehrends:
I'll have to look at that. The generated HTML is pretty complex and also seems to reference some external files. It may be that it can't reference that.

fingolfin:
Which browser's are you each using?

ThomasBreuer:
@fingolfin I am using Firefox 76.0.1.

rbehrends:
@fingolfin I doubt it's a browser issue, as I have the problem in both Firefox and Chrome when viewing it on GitHub or in nbviewer, but it works at least partly if I view it in a local Jupyter notebook server. There's something going wrong with server-side conversion on GitHub/nbviewer (the notebook is a JSON file that needs to be converted to HTML).

On closer inspection, part of the problem may be that the visualization produces a form with sliders and checkboxes that GitHub/nbviewer cannot handle. Even when I open it locally, I only get the sliders and checkboxes and have to rerun the notebook for the actual visualization. I suspect that that may go beyond what we can display without having live access to a Jupyter instance.

ThomasBreuer:
For the moment (that is, until we know a safe way to deal with such output not generated by Julia), it is probably the best solution to omit the graph visualization.

rbehrends:
The notebook has been updated; I've also changed the image to one for Oscar (which may also be a good choice for some other notebooks that show interactions between the subsystems).

rfourquet (approved):

PR #116, opened 2020-05-16
examples: hide by default instructions to download notebooks

rfourquet:
These instructions are quite noisy and steal the attention away from the notebooks themselves.
So let's hide them by default, using html's `<detail>`'s feature here.

Feedback welcome for the css style: either no css, the same as that used in the "installation" page, or something else. Preview at https://rfourquet.github.io/oscar-website/example/ for the first two options.

### DIFF
--- a/example.html
+++ b/example.html
@@ -4,32 +4,67 @@ title: OSCAR examples
 ---
 
 Clicking on the example will open a static version of the example, powered by <a href="https://nbviewer.jupyter.org/">nbviewer</a>.
-To interact with a "live" version, it is possible from there to download the notebook and run it locally:
-<ul>
-  <li>open the context menu (via right-click on Linux and Windows, or ctrl-click on macOS)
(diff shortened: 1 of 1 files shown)


mohamed-barakat:
I like the boxed one. It is clearer that there is there something to click-open. The plain version just looks like a bulleted item.

fingolfin:
I also prefer the boxed one, for the reasons @mohamed-barakat mentioned. There is one thing I dislike about it, though: It is indented compared to the stuff before and after it; yet it is also much wider than what comes before and after it. This looks really odd. I think it should be aligned visually with the boxes following it. *That said*, of course the stuff coming after it should be completely redesigned soon, too, so we might want to ignore this for now and just merge this as-is (well, with the duplicates removed, of course)

rfourquet:
Yes I totally agree the alignment looks odd, I will try to tweak it now and merge it today.

rfourquet:
I updated with the same alignment as the boxes below.

Issue #92, opened 2020-04-22
Come up with a nicer way to display the examples


fingolfin:
The examples display on https://oscar.computeralgebra.de/example/ does not look so nice. This needs to be improved; and once we have more examples, we must also think about how to make this more accessible. Some thoughts and ideas (to be refined, i.e. none of this is set in stone):
- perhaps the examples could be narrower, but then two side by side? (Perhaps a responsive design with "cards" a bit like on <https://oscar-system.github.io/oscar-ci-status/index.html> could be used?)
- assign "keywords" or "tags" to each example, e.g.:  "GAP, Singular, tropical matroids" which give a rough indication which technologies  and mathematical areas are in the example (@rfourquet shall provide the setup, but others can and should help with determining what tags are appropriate)
- once there are many examples (I wish...) we can think about allowing to filter the displayed examples by tag (e.g. "only show examples with tags "GAP" and "tropical" or so)
- .. more? Not sure

fingolfin:
@rfourquet This will become more pressing as new examples are added. Can you please take a look now?

fingolfin:
@rfourquet I really think we should make each example "card" narrower, and then show multiple of them next to each other -- this could be done dynamically. Let me again point at https://oscar-system.github.io/oscar-ci-status/ which I created in less than an hour of work. The central aspect is https://github.com/oscar-system/oscar-ci-status/blob/master/assets/css/cards.css which has a comment saying where I stole the inspiration. The key is the `grid-template-columns` css selector, and once understood, it's pretty easy to adjust.

rfourquet:
Sure I can try to do that now. I just didn't have a personal insight that this would look better.

PR #117, opened 2020-05-16
add examples filtering feature

rfourquet:
This tries to address the "filtering examples" part of #92.

As we have a static website (i.e. without javascript), I created a separate page for each category. If we want more refined filtering (e.g. according to two or more keywords), I'm afraid we would need to use javascript, but I mostly don't know about this, so better ideas are welcome.

Others will need to chime in order to decide which keywords we use, for now I simply used these implied by the logos. There is a "main example page" (with keyword "All") containing all examples, and subpages with filtered examples. 

Preview at https://rfourquet.github.io/oscar-website/exampleNEW/.


fingolfin:
That's a good start, IMHO. But it should be changed so that the currently selected "filter" is highlighted (e.g. instead of being a link, it should be bold). We use a similar trick for the sidebar navigation, see `_includes/sidebar.html` (this is a fairly standard technique, so google should reveal plenty more examples how to do it in Jekyll).

rfourquet:
> it should be changed so that the currently selected "filter" is highlighted 

Yes this was my plan, but I wanted to have feedback on the approach first before finalizing.

mohamed-barakat:
Logo_CAP could be renamed into CAP.

rfourquet:
> Logo_CAP could be renamed into CAP.

Yes this will of course be done. 

The current state of this PR is really not finalized, but before proceeding I wanted to have feedback on this filtering according to a category which can be mapped to "thumbnails".
I find it relatively intuitive for the user. But without javascript it's difficult to filter according to more than one categorization/keyword, so a decision needs to be made on which one.

mohamed-barakat:
I think it is enough for now given the small about of notebooks so far.

rfourquet:
I updated and finalized, preview now at https://rfourquet.github.io/oscar-website/example/.
As more notebooks are converted to using `Oscar` only, I don't know how this filtering will evolve, but this "provides the setup".

[EDIT: this is rebased on top of the "tiled examples" PR, which should be merged before this one]

rfourquet:
I will merge later today if no objections.

PR #125, opened 2020-05-20
add "Tropicalization notebook"

rfourquet:
cc @micjoswig

### DIFF
--- a/_data/examples.yml
+++ b/_data/examples.yml
@@ -102,3 +102,10 @@
   author: Clara Petroll
   thumbnail: oscar.png
   language: julia
+
+- title: "Tropicalization of a p-adic surface"
+  repository: micjoswig/oscar-notebooks
+  filename: patchworking/Tropicalization
+  author: Michael Joswig
+  thumbnail: oscar.png
+  language: julia


_data/examples.yml
```diff
@@ -102,3 +102,10 @@
   author: Clara Petroll
   thumbnail: oscar.png
   language: julia
+
+- title: "Tropicalization of a p-adic surface"
+  repository: micjoswig/oscar-notebooks
+  filename: patchworking/Tropicalization
+  author: Michael Joswig
+  thumbnail: oscar.png
```
rfourquet, line 110: Should this use instead the `polymake.png` logo?

fingolfin: This does `using Oscar` and does not just use polymake, so I think it'd be fine to have it use the oscar logo. If @micjoswig changes his mind, it'll be easy enough to change it later. So just merge it as-is for now?

fingolfin (approved):

_data/examples.yml
rfourquet: I was thinking as much, I will merge later today.

PR #126, opened 2020-05-20
Tweak example cards

fingolfin:
Based on a PI discussion we had, this PR makes a few changes:

- drop useless "language" line
- drop useless notebook badge
- make everything take up fewer space
- render the author(s) differently from the title

TODO:
- for the "G-Vectors" example, the author name appears in the same line as the title; it should not
- perhaps the content of the bars could be vertically centered?
- test this with more browsers.

@rfourquet I don't have time to complete this, so please take over from here and try to resolve the above TODOs. Feel free to push to this PR, or create a new one based on what I did here.


rfourquet:
> drop useless "language" line

Very nice, I've been meaning to do the same for a while.

> drop useless notebook badge

Sounds good too. I believe that the badge was meant initially to link directly to a live version of the notebook, but this was disabled. If needed, when we can provide live notebooks, we can always re-enable these badges. 

I will try to resolve the TODOs.

rfourquet:
So tweaked a bit more CSS and tested on few browsers where it seems to work (brave, firefox, luakit, qutebrowser). 

I don't have a strong opinion concerning the alignment, so I put them all at https://rfourquet.github.io/oscar-website/example/ to get feed-back (if all the blocks show a left-alignment like it did to me, try a moment later and "hard-refresh" the page).

Sorry it took so long...

mohamed-barakat:
Thank you very much. I like the left alignment.

fingolfin:
Left or right alignment are both fine by me, I don't really like the one with centering. Since Mohamed likes the left alignment, I'd be happy to just go with that for now.

I also think that vertical centering of the icons might look nicer. But that good be dealt with in a later PR.

rfourquet:
Ok thank you both for the feed-back, I will then go with left-aligned text (which I also slightly prefer).

> I also think that vertical centering of the icons might look nicer

Yes I tried to do that without success so far. I saw online a potential solution involving "flex". Or an alternative might be to have all the icons have the same height and thus different width, but making sure that the alignment of the text on the right is not affected by that.
I will look into that.

Issue #113, opened 2020-05-13
Sort the examples differently, perhaps group them into categories


fingolfin:
The examples should come in a different order: first there should be basic examples ; then more advanced ones.  We could also group things into categories.

To decide what is basic and what is not, as a first approximation, you could do this: Open the notebook; if you immediately understand what's going on (and it's not just because it's your personal area of expertise), it's basic; if not, then it is probably not (it might be basic *in its area*, but not "basic" in a global context).

We can of course refine and change later on what is "basic" and what is "advanced", resp. what is which category; the important bit here is to create infrastructure so that we can group things this way, and to have an initial on this.

Related to issue #92

PR #128, opened 2020-06-02
classify examples according to "level": intro, etc.

rfourquet:
This is a first step to address #113: present first intro-level examples, then more advanced ones. Currently I set up 3 categories: 1-introduction, 2-intermediate, 3-advanced. I definitely need help for classifying the notebooks, as I'm not familiar with many of the math topics used in them. I mostly used "intermediate" for when I'm less sure, but 2 categories might be enough. Or the level could be a simple integer, say from 1 to 5, set by the author.

Also, do we want visual indication of the level besides the order of presentation? If so, should it be written in the cards, or have the cards separated in groups?

### DIFF
--- a/_data/examples.yml
+++ b/_data/examples.yml
@@ -11,6 +11,7 @@
   author: Michael Joswig
   thumbnail: polymake.png
   language: julia
+  level: 2-intermediate
 
 - title: "Recognizing Manifolds"
   repository: micjoswig/oscar-notebooks
@@ -18,6 +19,7 @@
   author: Michael Joswig
   thumbnail: polymake.png
   language: julia
+  level: 2-intermediate
 
 - title: Singular.jl example
   repository: oscar-system/OSCARBinder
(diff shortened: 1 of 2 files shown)


fingolfin:
Ultimately I'd like to be able to "filter" the groups. I.e. at the top of the interface, have some radio buttons or  which the user can to select which e.g. topics / tags / levels are shown and which not. The user might also be able to affect the sorting this way.

I never did much JavaScript but nevertheless was able to cobble together http://groups.quendi.de with just some googling (I mean, not that it looks great or is a great example, all I want to say: it isn't *that* hard. Main code in the HTML file and in https://github.com/fingolfin/smallgroups-web/blob/gh-pages/js/code.js)

HereAround:
Generally speaking, I like the idea.

However, this is hopelessly outdated (no action since July, 2020) and we are in the process of a massive overhaul of the examples/tutorials anyways. As just discussed with @fingolfin , it therefore seems best to close this PR.

Still - thank you for looking into this @rfourquet .

PR #130, opened 2020-07-17
A divisionally free not inductively free arrangement with a minimal number of hyperplanes

mohamed-barakat:


rfourquet:
Unfortunately the title is longer than our layout logic can currently handle (the text overflows out of the "card"/frame). Cf. https://rfourquet.github.io/oscar-website/example/CAP/ (even two lines get out of the card when the window is made narrower). 

Hopefully this will get fixed rapidly.

mohamed-barakat:
> Hopefully this will get fixed rapidly.

Do you suggest to change the title or will the layout be changed?

rfourquet:
> Do you suggest to change the title or will the layout be changed?

No no, I was suggesting that hopefully I'm able to fix the layout rapidly (e.g. on Monday), possibly by just increasing the size of all cards. Sorry for my lack of clarity.

mohamed-barakat:
Thank you very much, absolutely no problem.

fingolfin (approved):
With PR #131 merged this should be ready for merging, too

fingolfin:
Unfortunately browser caching of the CSS bites us again. I had to hold "shift" while choosing to reload the page in order to force my browser to do a full reload, including the CSS; with a plain reload, the new card for this example was indeed overfull. (BTW, I used Cmd-Shift-R in Firefox all the time, but in Safari, it annoyingly triggers "reader mode" now; one has to click the reload button and/or use the reload menu item, while holding shift. Stupid Apple).

Anyway, I wonder if there is something we can do to make browsers better recognize the CSS changed...

fingolfin:
Relevant StackOverflow post: https://stackoverflow.com/questions/118884/how-to-force-the-browser-to-reload-cached-css-js-files

PR #131, opened 2020-07-20
examples: make cards higher (allow 4 lines per cards)

rfourquet:
To accomodate upcoming notebook example (cf. #130).

### DIFF
--- a/public/css/hyde.css
+++ b/public/css/hyde.css
@@ -401,17 +401,18 @@ a.sidebar-nav-item:focus {
     margin: 0 auto;
     display: grid;
     grid-gap: 1rem;
-    grid-template-columns: repeat(auto-fit, minmax(16rem, 1fr));
+    grid-template-columns: repeat(auto-fit, minmax(20rem, 1fr)); /* use 16rem for 3 lines per card */
 }
 
 .item_example_card {
   background-color: #e4e4e7;
   padding: 10px;
   border-radius: 5px;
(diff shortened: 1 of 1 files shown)


fingolfin:
Can we see a preview somewhere?

rfourquet:
https://rfourquet.github.io/oscar-website/example/

(this preview also includes PR 130, i.e. the new notebook which has longer tittle than usual "A divisionally free not inductively free arrangement with a minimal number of hyperplanes")

mohamed-barakat (approved):

mohamed-barakat:
Thank you very much. Looks nice.

fingolfin:
Interesting. For me, this PR adds a lot of vertical whitespace in each card, but it doesn't really look to me as if it was needed for PR #130. Perhaps depends on which browser is used and where?

Screenshot 2020-07-20 at 16 18 10

fingolfin:
(and yes, I did try resizing it back and forth)

rfourquet:
In your picture, it looks like there is only one column of cards; when you have two columns, do you observe the same effect (unnecessary vertical space) ? 

For me, on the few browsers I tried, it looks like that:

pr130

But otherwise, when only one column is displayed, there is indeed more vertical space than necessary. I don't whether it's possible to change that.

fingolfin:
OK, I see it now! Fine by me then!

fingolfin (approved):

Issue #112, opened 2020-05-13
Make sure all example notebooks start with `using Oscar` (or have a very good reason not to)


fingolfin:
Please go through all example notebooks; for any that is not `using Oscar`, please contact the respective authors and work with them to get the notebooks update. (For the one authored by Sebastian Gutsche, I think we need somebody else to take over as "maintainer", as Sebastian left to academia).

Such a conversion can use the fact that `Oscar` imports the other cornerstones, so usually this should be a relatively minor change.

mohamed-barakat:
Just for the record: The banner displayed when `using HomalgProject` explicitly mentions Oscar and the used cornerstones.

rfourquet:
Could you explain the rationale? Is it to make the experience of the user simpler? To promote `Oscar` rather the components?  

What to do for example about ["Example AbstractAlgebra.jl"](https://nbviewer.jupyter.org/github/oscar-system/OSCARBinder/blob/master/AbstractAlgebra.ipynb) : its purpose seems to show few things available in this base package; but replacing `using AbstractAlgebra` by `using Oscar` will create `Nemo` objects instead (and the title would have to be updated)... so I'm not sure what to ask the author for.

fingolfin:
The rationale is that this is the website for Oscar, and so the examples should be for Oscar. 

And we also will have to remove, change or replace many of the examples. Changing them all to start with `using Oscar` is in so far just a first step; the real goal is to have actual examples for "Oscar".

rfourquet:
Sorry for the delay. Here the list of notebooks without `using Oscar`:

* `Hecke`
  * [ ] [Intro Number Fields: Towers](https://github.com/oscar-system/OSCARBinder/blob/master/Hecke.ipynb) (@fieker)
* `Polymake` (@micjoswig)  
  * [ ] [G-vectors](https://github.com/micjoswig/oscar-notebooks/blob/master/polymake/g-vectors_of_random_simplicial_6-polytopes.ipynb)
  * [ ] [Recognizing Manifolds](https://github.com/micjoswig/oscar-notebooks/blob/master/polymake/K3-16.ipynb)

* CAP (3 notebooks, @mohamed-barakat): do they need to be updated, given that

> The banner displayed when using HomalgProject explicitly mentions Oscar and the used cornerstones.

?

* Specific packages examples:
[...]

PR #137, opened 2020-08-13
Added two notebooks regarding homolgy computations with Polymake

IvanSpirandelli:


rfourquet:
Thanks. Note that @fingolfin has requested that notebooks should generally try to have `using Oscar` at the top rather than using "subpackages" like `Polymake` ("sub" in the sense that they are imported by `Oscar`), cf. #112.

PR #141, opened 2020-09-04
SymmetricAlgebraInternalToS4rep

mohamed-barakat:
Computations in the symmetric algebra internal to the tensor category of finite dimensional representations of Sym(4)


rfourquet:
The title is longer than what would currently fit in the "example cards", I will need to adjust the CSS.

mohamed-barakat:
Thank @rfourquet 

@ThomasBreuer: Dear Thomas, these are the sort of computations I told you about in Kaiserslautern.

PR #148, opened 2021-01-15
Update Julia install instructions & Oscar banner

fingolfin:
Many people are using Ubuntu, and install Julia via `apt install julia`. But
that yields a semi-broken experience -- e.g. for GAP.jl to work, one at the
very also needs to install the `libjulia-dev` package. So let's mention
alternatives for installation and warn about using distro package managers.

Also don't emphasise Julia version 1.3.1 too much, we don't want people who
already have a newer Julia to think they ought to downgrade.

Finally, update the Oscar banner, so that people who by a mishap end up with
an outdated Oscar installed at least have a chance to notice that it is
outdated ("Huh, the website says 0.5.0, why do I have 0.2.0?")

### DIFF
--- a/install.md
+++ b/install.md
@@ -133,10 +133,21 @@ To start bash in a later session, just search for <q>bash</q>.
 
 </div>
 
-## Step 2: Install Julia 1.3.1 or newer
+## Step 2: Install Julia
 
-Install Julia 1.3.1 or newer [by downloading it from the Julia homepage](https://julialang.org/downloads/),
+OSCAR requires at least Julia 1.3.1, but we recommend running it with the latest stable Julia release,
+which is 1.5.3 at the time this is written.
+
+There are several ways to install Julia:
+
+1. [By downloading it from the Julia homepage](https://julialang.org/downloads/),
 and following their [platform specific instructions](https://julialang.org/downloads/platform/).
+
+2. On macOS, you can also install it via [Homebrew](https://brew.sh): `brew install julia` 
+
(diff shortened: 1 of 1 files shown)


install.md
```diff
@@ -133,11 +133,22 @@ To start bash in a later session, just search for <q>bash</q>.
...
+which is 1.5.3 at the time this is written.
+
+There are several ways to install Julia:
+
+1. [By downloading it from the Julia homepage](https://julialang.org/downloads/),
 and following their [platform specific instructions](https://julialang.org/downloads/platform/).
 
+2. On macOS, you can also install it via [Homebrew](https://brew.sh): `brew install julia` 
+
+3. The [JILL](https://github.com/johnnychen94/jill.py) project is a python package which allows installing and updating Julia on Linux, macOS, Windows -- this is in particular handy for experienced users who may want to install multiple Julia versions in parallel; but also for beginners it can be convenient as it allows updating the installed Julia version quite easily.
```
```suggestion
3. The [JILL](https://github.com/johnnychen94/jill.py) project is a python package which allows installing and updating Julia on Linux, macOS -- this is in particular handy for experienced users who may want to install multiple Julia versions in parallel; but also for beginners it can be convenient as it allows updating the installed Julia version quite easily. (JILL also supports Windows, but as explained in step 1 above, OSCAR does not work under Windows directly, but rather requires the Windows Subsystem for Linux (WSL).)
```

rfourquet (approved):

PR #155, opened 2021-01-27
Meeting March 2021

fingolfin:
@fieker please have a look if this seems roughly sensible. Then we can put it online, and advertise it.


fieker:
On Wed, Jan 27, 2021 at 02:54:06AM -0800, Max Horn wrote:
> @fieker please have a look if this seems roughly sensible. Then we can put it online, and advertise it.
The monday programm is not listing the discussion....

PR #160, opened 2021-02-17
add some IJulia tips

rfourquet:
cc @micjoswig


thofma (approved):

thofma:
@fingolfin

fingolfin (approved):
I can't really comment much, I think it was @micjoswig who really wanted this addition, so he should say if this matches his vision; but what is here looks like a good start to me.

So I'd say if Michael doesn't reply today, just merge it?

PR #189, opened 2021-09-07
fix test for julia exercise 7

lgoettgens:
The original code
```julia
g = cached(+)
println(g(1, 1)) == 1 + 1
```
compares `nothing` (the return value of `println`) to 2.


fingolfin (approved):
Thank you!

PR #204, opened 2021-09-08
Fix Gram matrix

thofma:


fingolfin:
This way the matrix appears but as a 1x9 row vector

thofma:
1x7 row vector ...

PR #213, opened 2021-09-14
add info about a system-wide Oscar installation

ThomasBreuer:


install.md
```diff
@@ -230,3 +230,113 @@ the [trouble shooting page](https://julialang.github.io/IJulia.jl/stable/manual/
...
+
+# Let Julia install and precompile the packages.
+${julia_for_oscar} -e 'using Pkg; \
+Pkg.add("Oscar"); using Oscar; \
+Pkg.add("GAP"); using GAP; \
+Pkg.add("Nemo"); using Nemo; \
+Pkg.add("Hecke"); using Hecke; \
+Pkg.add("Singular"); using Singular; \
+Pkg.add("Polymake"); using Polymake; \
+Pkg.add("AbstractAlgebra"); using AbstractAlgebra; \
```
fingolfin: Why is it necessary to install all of these; doesn't just Oscar suffice?

install.md
```diff
@@ -230,3 +230,113 @@ the [trouble shooting page](https://julialang.github.io/IJulia.jl/stable/manual/
...
+# Specify the intended location of the central Oscar installation.
+central_depot=/users/oscar/JULIA_DEPOT
+
+# Set the Julia variables that control the location of packages.
+export JULIA_DEPOT_PATH=${central_depot}
+export JULIA_LOAD_PATH="@:@v#.#:@stdlib:${central_depot}/environments/globalenv"
+
+# Let Julia update and precompile the installed packages.
+${julia_for_oscar} -e 'using Pkg; \
+Pkg.update(); \
```
fingolfin: Won't this easily get stuck, as we saw last week, because all those dependencies are also added to the active environment?

fingolfin: Perhaps it would be safer to just delete and recreate the environment? This way, more code could be shared with the first script, too

fingolfin (commented):
Thanks! I'll properly look at this tomorrow.
 Is there a preview somewhere?

UPDATE: I pushed a preview to https://fingolfin.github.io/oscar-website/install/#system-wide-installation-of-oscar

install.md
ThomasBreuer: If only Oscar is added to the environment then the other packages are installed physically but they are not added to the `Project.toml`. A user calling `using GAP` (for example) will then get the message that first one has to call `Pkg.add("GAP")`, and trying this will result in the error message that the user has no write permission for the central installation.

install.md
ThomasBreuer: Should we then say that updating means to remove the installation and to install it again, and thus no dedicated script for updating is needed at all?

fingolfin: That would potentially break things for people who are running Julia/Oscar *right now*. This is just about the environment.

install.md
fingolfin: I think I don't understand how this depot is intended to be used. I thought it was mainly about caching all the large artifacts and package files, so users don't need to have them in their home directories (which might also be on NFS)

But it seem you are suggesting everyone should be using *that same environment*? I would have thought everyone is still expect to at least once `Pkg.add("Oscar")`  (or, of course, `Pkg.add("GAP")`). But OK, I guess that doesn't mesh well with "having Oscar up-to-date all the time". On the other hand, *that* doesn't mesh well with "please don't break my working Oscar behind my back"... Hrm

install.md
ThomasBreuer: O.k., if we remove the `Project.toml` before adding the packages then we also need just one script for installing and updating.

install.md
ThomasBreuer: If the idea is to share just the artifacts then we can provide a recommended `Overrides.toml` file that points to directories in the `JULIA_DEPOT/artifacts` in question.
In fact, the two approaches do not exclude each other.

install.md
ThomasBreuer: @fingolfin I have updated the pull request in order to follow your idea.

install.md
fingolfin: But Julia allows for a stack of active depots, see https://docs.julialang.org/en/v1/base/constants/#Base.DEPOT_PATH -- note that` /usr/local/share/julia` is on that default list. So if one just puts artifacts in there, they should be loadable by any Julia user on that system without any additional effort.

ThomasBreuer: @fingolfin I understand your idea as follows: First make an Oscar installation. Then take the artifacts that have been installed during this installation, and move them to the "architecture-specific shared system directory" that can be found in `DEPOT_PATH`. This way one need not provide an `Overrides.toml`. For updating an existing "shared Oscar installation", one repeats this process. Is this interpretation correct?

fingolfin:
I am sorry for being slow on responding to this. I'll add more remarks later. In the meantime, I just came across https://docs.julialang.org/en/v1/manual/faq/#Computing-cluster and perhaps we should also include this advice, and/or link to it.

install.md
ThomasBreuer: I have made more experiments.

Providing the necessary artifacts inside the "architecture-specific shared system directory" as sketched above works.
(Well, some artifact seems to be downloaded later on, and hence ends up in the user's `.julia/artifacts` directory; I think this is not desirable.)

However, I have not found a way to make the idea to share artifacts compatible with the idea to have a complete installation of Oscar that is shared between users. (I tried to use the artifacts inside the "architecture-specific shared system directory" by the central installation, but some packages try to find the artifacts in the directory of the central installation.)

ThomasBreuer:
@fingolfin Here is the traceback of the error I had mentioned: After installing Oscar with the script, I moved the artifacts as follows.
```
system_depot=$(${julia_for_oscar} --startup-file=no -e 'println(DEPOT_PATH[2])')
if ! test -d "${system_depot}/artifacts"; then
  mkdir -p "${system_depot}/artifacts"
fi
mv ${central_depot}/artifacts/* ${system_depot}/artifacts
```
Then setting
```
export JULIA_DEPOT_PATH=$HOME/.julia:${system_depot}:${central_depot}
```
and starting Julia yields
```
julia> DEPOT_PATH
3-element Vector{String}:
 "/users/oscar/.julia"
 "/users/oscar/julia-1.6.2/local/share/julia"
 "/users/oscar/JULIA_DEPOT"

julia> using Oscar
ERROR: InitError: could not load library "/users/oscar/JULIA_DEPOT/artifacts/16e1de4679fb8520a8af4e6831c7c8e9893d18b4/lib/libcxxwrap_julia_stl.so"
/users/oscar/JULIA_DEPOT/artifacts/16e1de4679fb8520a8af4e6831c7c8e9893d18b4/lib/libcxxwrap_julia_stl.so: cannot open shared object file: No such file or directory
Stacktrace:
  [1] dlopen(s::String, flags::UInt32; throw_error::Bool)
    @ Base.Libc.Libdl ./libdl.jl:114
  [2] dlopen(s::String, flags::UInt32)
    @ Base.Libc.Libdl ./libdl.jl:114
  [3] initialize_julia_module(mod::Module)
    @ CxxWrap.CxxWrapCore ~/JULIA_DEPOT/packages/CxxWrap/OcN1Z/src/CxxWrap.jl:412
  [4] __init__()
    @ CxxWrap.StdLib ~/JULIA_DEPOT/packages/CxxWrap/OcN1Z/src/StdLib.jl:20
  [5] _include_from_serialized(path::String, depmods::Vector{Any})
    @ Base ./loading.jl:696
  [6] _require_search_from_serialized(pkg::Base.PkgId, sourcepath::String)
    @ Base ./loading.jl:782
[...]
```
The artifact can be found in `DEPOT_PATH[2]` but somehow it is expected in `DEPOT_PATH[3]`.
Is it perhaps not allowed to move artifacts after the installation of packages that rely on them?
(I had expected that one can prescribe a path for the artifacts to be installed, but I did not find a way to do this.)

fingolfin:
@ThomasBreuer can you please update it so it merges again?

@hannes14 can you have a quick look at this, please?

hannes14 (approved):
polymake requires (sometimes) a directory writable by the user.
After line 303 add:
export POLYMAKE_USER_DIR=$HOME/.julia/polymake_user

Line 306 can be changed to:
${julia_for_oscar} --load ${central_depot}/julia_load_initial.jl $*
to catch additional arguments to the script

ThomasBreuer:
Thanks @hannes14 !

ThomasBreuer:
@fingolfin I think that my proposal does not work at all.

For example, if one creates such a system-wide installation, starts Julia, and then calls `using Oscar; Oscar.allow_unicode(true)` then one runs into an error because `set_preferences!` (from Preferences.jl) tries to write a file `LocalPreferences.toml` in the current environment. However, this environment is common for several users. (Yes, this is apparently a bad idea.)

The idea was to have one installation which is precompiled and thus can be regarded as read-only and thus can be shared by several users. But this is apparently not true. Even if it would be true now, we can expect that later on some functions will try to write some files in this installation.

Perhaps we cannot achieve much more than sharing the artifacts.

How do people who (for example) are running computer labs for many students solve the problem that Julia is so hungry for disk space?

fingolfin:
Julia supports multiple active environments, and multiple active depots. I thought this would work by reading from all as necessary, but writing only happens in the "top" environment / depot? I understand @ThomasBreuer as saying this is not the case in Preferences?

Note: We should share depots, but not environments.

Even if there is a problem, we can for now just warn users about this, and move on. But of course there might also be a bug in `Preferences.jl` (then we should report it), or a mistake in how we do things (then we should fix it)

fingolfin:
@hannes14 wrote:
> polymake requires (sometimes) a directory writable by the user.

Is this still the case, doesn't Polymake.jl now make Polymake use a Julia scratch space? (@benlorenz ?)

benlorenz:
> @hannes14 wrote:
> 
> > polymake requires (sometimes) a directory writable by the user.
> 
> Is this still the case, doesn't Polymake.jl now make Polymake use a Julia scratch space? (@benlorenz ?)

Polymake.jl will put everything in a scratch space, so please don't set `POLYMAKE_USER_DIR`. The scratchspace is regenerated during `__init__` so this should be fine regarding paths and precompilation.

install.md
```diff
@@ -198,3 +198,96 @@ If you are using OSCAR in the Windows Subsystem for Linux, you will require a br
...
+# Specify the intended location of the central Oscar installation.
+central_depot=/users/oscar/JULIA_DEPOT
+
+# We will store the necessary artifacts in the
+# "architecture dependent location" in Julia's default DEPOT_PATH.
+# First determine this path and create the directory if necessary.
+system_depot=$(${julia_for_oscar} --startup-file=no -e 'println(DEPOT_PATH[2])')
+if ! test -d "${system_depot}/artifacts"; then
+  mkdir -p "${system_depot}/artifacts"
+fi
```
```suggestion
mkdir -p "${system_depot}/artifacts"
```
I think `mkdir -p` already is idempotent, so there should be no need for a separate `test`?

install.md
```diff
@@ -198,3 +198,96 @@ If you are using OSCAR in the Windows Subsystem for Linux, you will require a br
...
+#!/bin/bash
+
+# Specify the intended version of Julia.
+julia_for_oscar=/users/oscar/julia-1.8.5/bin/julia
+
+# Specify the intended location of the central Oscar installation.
+central_depot=/users/oscar/JULIA_DEPOT
+
+# Admit the user's own depot path and the path with the central installation.
+export JULIA_DEPOT_PATH=$HOME/.julia:${central_depot}
```
fingolfin: I think this is too much as it also remove the system depot from the list of depots.

You can also prepend/append to the depot list, though. Here I would suggest:
```suggestion
export JULIA_DEPOT_PATH=:${central_depot}
```
Then the "central depot" is last in the list of all depots

install.md
```diff
@@ -198,3 +198,96 @@ If you are using OSCAR in the Windows Subsystem for Linux, you will require a br
...
+Pkg.add("Hecke"); \
+Pkg.add("Singular"); \
+Pkg.add("Polymake"); \
+Pkg.add("AbstractAlgebra"); \
+Pkg.instantiate(); \
+end; \
+exit();'
+
+# Write a file that will be loaded by Julia on startup.
+echo 'using Pkg; Pkg.activate("'${central_depot}'/environments/v" * join(split(string(VERSION), ".")[1:2], "."))' > ${central_depot}/julia_load_initial.jl
```
fingolfin: We may not need this at all if we use stacked environments right

install.md
```diff
@@ -198,3 +198,96 @@ If you are using OSCAR in the Windows Subsystem for Linux, you will require a br
...
+# First determine this path and create the directory if necessary.
+system_depot=$(${julia_for_oscar} --startup-file=no -e 'println(DEPOT_PATH[2])')
+if ! test -d "${system_depot}/artifacts"; then
+  mkdir -p "${system_depot}/artifacts"
+fi
+
+# Set the Julia variables that control the location of packages.
+# (Do not admit the current user's own depot path.)
+export JULIA_DEPOT_PATH=${central_depot}:${system_depot}
+export JULIA_LOAD_PATH="@:@v#.#:@stdlib:${central_depot}/environments/globalenv"
```
fingolfin: I think the name of the global environment `globalenv` should involve the Julia version. I think we can achieve that here by just renaming it to e.g `@v#.#-oscar` or something like that.

install.md
```diff
@@ -198,3 +198,96 @@ If you are using OSCAR in the Windows Subsystem for Linux, you will require a br
...
+
+# Specify the intended version of Julia.
+julia_for_oscar=/users/oscar/julia-1.8.5/bin/julia
+
+# Specify the intended location of the central Oscar installation.
+central_depot=/users/oscar/JULIA_DEPOT
+
+# Admit the user's own depot path and the path with the central installation.
+export JULIA_DEPOT_PATH=$HOME/.julia:${central_depot}
+export JULIA_LOAD_PATH="@:@v#.#:@stdlib:${central_depot}/environments/globalenv"
```
fingolfin: Ah and here this trick with "appending" also work:
```suggestion
export JULIA_LOAD_PATH=":@v#.#-oscar"
```
Then I get, after starting julia:
```
julia> LOAD_PATH
4-element Vector{String}:
 "@"
 "@v#.#"
 "@stdlib"
 "@v#.#-oscar"
```

Oh and for the depot I got the following (with `central_depot=/julia_depot`)
```
julia> DEPOT_PATH
4-element Vector{String}:
 "/root/.julia"
 "/usr/local/julia/local/share/julia"
 "/usr/local/julia/share/julia"
 "/julia_depot"
```

fingolfin: And this worked just fine:
```
julia> using Singular
Singular.jl, based on
                     SINGULAR                               /
 A Computer Algebra System for Polynomial Computations     /  Singular.jl: 0.18.13
                                                         0<   Singular   : 4.3.2p7
 by: W. Decker, G.-M. Greuel, G. Pfister, H. Schoenemann   \
FB Mathematik der Universitaet, D-67653 Kaiserslautern      \
```
Even though `Pkg.status()` reports no packages are installed... That's a bit confusing for users, of course (and might make debugging issues a bit harder for them: "which version of AbstractAlgebra are you using?" is harder to answer then, maybe...

install.md
```diff
@@ -198,3 +198,96 @@ If you are using OSCAR in the Windows Subsystem for Linux, you will require a br
...
+# Set the Julia variables that control the location of packages.
+# (Do not admit the current user's own depot path.)
+export JULIA_DEPOT_PATH=${central_depot}:${system_depot}
+export JULIA_LOAD_PATH="@:@v#.#:@stdlib:${central_depot}/environments/globalenv"
+
+# Clean the environment, such that the already centrally installed packages
+# get replaced by newer versions if necessary.
+# (This is safer than calling `Pkg.update()` in Julia.)
+# Then let Julia install and precompile the packages.
+${julia_for_oscar} -e 'using Pkg, Artifacts; \
```
```suggestion
${julia_for_oscar} --project=@v#.#-oscar -e 'using Pkg, Artifacts; \
```
To ensure the packages are added to the "right" environment

ThomasBreuer:
Thanks for the comments.
Now the scripts use named environments, which makes them simpler.

The technical problem is still the same:

If Julia is called with the project `@v#.#-oscar` then the environment is found in the depot of the central installation, and thus is the same for all users, and `Oscar.allow_unicode(true)` (or using `Oscar.with_unicode() do ...`, as some testfiles do) will try to write the file `LocalPreferences.toml` in the directory of `Base.active_project()` and hence run into an error.
(see JuliaPackaging/Preferences.jl/issues/55.)

fingolfin:
That's why I deliberately did not propose to use `--project=@v#.#-oscar` for the user invoking Julia, but rather to use `JULIA_LOAD_PATH ` to insert `@v#.#-oscar` into the environment stack, but *below* `@v#.#` -- the latter will then be the active project and live in the users personal depot (so `~/.julia`). I.e.
```
export JULIA_LOAD_PATH=":@v#.#-oscar"
julia
```
inside the script

ThomasBreuer:
@fingolfin Ah, then I misunderstood your idea. I will try what you wrote now.

ThomasBreuer:
@fingolfin Yes, your proposal works in principle, and I have changed the script in question accordingly.

However, there are some subtle problems. (I have added a vague warning to the description.)
- Everything is fine if `~/.julia/environments/v1.8/Project.toml` does not contain the dependencies of Oscar: Start Julia with the script, call `using Oscar`, and Julia will not try to precompile anything.
- If `~/.julia/environments/v1.8/Project.toml` contains for example an entry about GAP.jl before using the system-wide installation for the first time, `using Oscar` triggers a precompilation.
- If one `Pkg.add`s AbstractAlgebra.jl and Nemo.jl after using the system-wide installation then they will get precompiled in the user's environment, and calling `using Oscar` afterwards runs into an error. (I do not understand what exactly goes wrong here.)
- It may happen that `~/.julia/environments/v1.8` does not exist. It will be created when the first package for Julia 1.8 gets `Pkg.add`ed, and users interested only in the system-wide installation of Oscar need not do this. In this situation, `Oscar.allow_unicode(true)` runs into an error. Perhaps Preferences.jl is not aware of the possibility that the directory is not available because it expects that at least Preferences.jl has been `Pkg.add`ed.

PR #216, opened 2021-10-10
Adding apt-upgrade to install.md

veni-vidi-code:
Reading through the install section on the website i 
Just calling apt-update does not actualy update the packages so i assume it is meant to also call apt-upgrade.

### DIFF
--- a/install.md
+++ b/install.md
@@ -49,6 +49,7 @@ Enter the following commands into a terminal (this will prompt for your password
 and requires that you have permissions to administer your computer).
 {% highlight bash %}
 sudo apt-get update
+sudo apt-get upgrade
 sudo apt-get install build-essential
 {% endhighlight %}
 </details>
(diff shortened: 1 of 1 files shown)


install.md
```diff
@@ -70,6 +71,7 @@ and requires that you have permissions to administer your computer).
 sudo apt-get install -y software-properties-common
 sudo add-apt-repository ppa:ubuntu-toolchain-r/test
 sudo apt update
+sudo apt-get upgrade
```
veni-vidi-code, line 74: im not sure if this is correct here since it just calls apt update and not apt-get update before

fingolfin: It should be apt-get there, too (well, it works either way, but....)

fingolfin:
The package is not meant to be updated. We just ensure it is installed.

Is there a concrete issue this PR is supposed to resolve? Or is it just out of principle?

PR #231, opened 2022-08-24
Update install instructions

thofma:
Do not recommend homebrew, since Oscar is likely to be broken.

Mention `juliaup` in addition to `JILL`.

CC: @benlorenz @hechtiderlachs


benlorenz:
We could also add macOS and homebrew to the Linux block in the warning above as the reasoning is very similar.

thofma:
Good point. How about that?

benlorenz (approved):
Very good, thanks!

PR #237, opened 2022-12-09
add data for a new notebook

ThomasBreuer:


fingolfin (approved):

_data/examples.yml
```diff
@@ -158,3 +158,10 @@
   author: Martin Schönert (translated to Oscar by Thomas Breuer)
   thumbnail: oscar.png
   language: julia
+
+- title: "Generation of sporadic simple groups by $\pi$- and $\pi'$-subgroups"
```
fingolfin, line 162: Actually this was not valid YAML, as it tries to interpret `\p` as a quoted character. This can be fixed by changing it to `\\p` but then the result still was not rendered as math by me. So I fixed this by using Unicode 𝜋

PR #248, opened 2023-03-15
Link preprint Toric Geometry in OSCAR

HereAround:


fingolfin:
Something is wrong with updating the website, and so your last few PRs are not yet online.

I recommend to check maybe 30 seconds after a merge of a PR was done here to see if the changes are online. If not, please open an issue.

I'll go and fix it again now, but I just did so a few days ago, so I need to figure out why it keeps breaking

fingolfin:
Root cause identified and fixed, it should work reliably again now. But of course never hurts to double check

PR #249, opened 2023-03-21
Instructions for notebooks in Windows Subsystem

YueRen:
By default Ubuntu installs browsers via snap which is disabled in Windows Subsystems. Worse, the Windows Subsystem will install the browser via snap, but starting a notebook from Julia will do nothing (not even return an error).

I added instructions on how to install a browser in the Windows Subsystem for Linux as a deb package. The method worked for me, let me know if there is a better solution.

### DIFF
--- a/install.md
+++ b/install.md
@@ -209,3 +209,8 @@ the [trouble shooting page](https://julialang.github.io/IJulia.jl/stable/manual/
 If you try to open an existing notebook (stored in a ".ipynb" file), it might refer to an
 older Julia version, resulting in a "Kernel error"; the solution is then to select
 a different kernel from the menu.
+
+If you are using OSCAR in the Windows Subsystem for Linux, you will require a browser 
+in your subsystem. This can be a probem as the default subsystem is Ubuntu and Ubuntu 
+installs browsers via snap which is disabled for subsystems. To circumvent this problem, 
+please see [how to install browsers via deb](https://www.omgubuntu.co.uk/2022/04/how-to-install-firefox-deb-apt-ubuntu-22-04).


fingolfin:
Alternatively, one could suggest users to configure their system to open the host system browser (so e.g. Edge). Instructions:

1. add `BROWSER='/mnt/c/Windows/explorer.exe'` to your environment. E.g. execute this `printf "\nexport BROWSER='/mnt/c/Windows/explorer.exe'" >> ~/.profile`

That's it. At least for me it works. Dunno if this will open e.g. Firefox or Chrome if the user chose those as their default browser?

Alternatively, one can use `wslu`:

1. install wslu (instruction: https://wslutiliti.es/wslu). For Ubuntu:
   ```
   sudo add-apt-repository ppa:wslutilities/wslu
   sudo apt update
   sudo apt install wslu
   ```
3. like above, edit `BROWSER=wslview`

YueRen:
I didn't know that `BROWSER='/mnt/c/Windows/explorer.exe` was an option, I think I prefer that solution (though with `MicrosoftEdge.exe` because `explorer.exe` opens the file browser).

YueRen:
~~@fingolfin I tried setting `BROWSER="MicrosoftEdge.exe"` on my system, but that didn't work for me. Microsoft Edge does start, but I get a "file not found" error.~~

YueRen:
I take back my last comment. My problem was caused because the path that is passed to Microsoft Edge is `home/ren/.local/share/jupyter/runtime/nbserver-279-open.html`. This path works for a browser installed in WSL, but not with a browser installed outside WSL (the correct path for the latter is `/mnt/wslg/distro/home/ren/.local/share/jupyter/runtime/nbserver-279-open.html`).

@fingolfin Was this a problem for your system as well? If yes, how did you solve it and make IJulia behave correctly?

fingolfin:
I did not try this, just tried opening other stuff. I can't access my Windows VM from outside the university, so I can't try alternatives right now -- it would be interesting to know if `wslview` from https://wslutiliti.es/wslu handles this better.

fingolfin:
I am going to merge this now; while perhaps we can do "even better" (whatever that might mean), the current PR is a clear improvement, and so let's get it out there!

Thanks again, @YueRen

PR #250, opened 2023-03-21
Updated installation instructions for WSL.

YueRen:
The old guide is out of date and installing WSL has become easier and can now be done in a single step.  For the ease of maintainance, I've replaced the old guide by the already existing link to the official installation instructions, and added a couple of hints on how to get started.

### DIFF
--- a/install.md
+++ b/install.md
@@ -22,11 +22,7 @@ terminal interface.
 <summary>
 Windows
 </summary>
-Please install Windows Subsystem for Linux (WSL) following the <a href="https://learn.microsoft.com/en-us/windows/wsl/install">official instructions</a>. You should now have an App "Ubuntu" in your start menu and a folder "Linux" on the left sidebar of your file explorer. You can now follow the instructions for <em><a href="#install-ubuntu">Ubuntu</a></em>. Please note: 
-<ol>
-<li> When you start Ubuntu, your terminal will be in the folder "Linux/Ubuntu/home/your_username>". Please avoid changing anything outside that folder unless you are absolutely sure.</li>
-<li> Your browser may be unable to download files directly to the "Linux" folder. Instead, download it to somewhere temporarily and move the file to "Linux" manually.</li>
-</ol>
+Please install Windows Subsystem for Linux (WSL) following the <a href="https://learn.microsoft.com/en-us/windows/wsl/install">official instructions</a>. You should now have an App "Ubuntu" in your start menu and a folder "Linux" on the left sidebar of your file explorer. You can now follow the instructions for <em><a href="#install-ubuntu">Ubuntu</a></em>. 
 </details>
 
 <details>
--- a/install.md
+++ b/install.md
@@ -22,7 +22,7 @@ terminal interface.
 <summary>
 Windows
 </summary>
-Please install Windows Subsystem for Linux (WSL) following the <a href="https://learn.microsoft.com/en-us/windows/wsl/install">official instructions</a>. You should now have an App "Ubuntu" in your start menu and a folder "Linux" on the left sidebar of your file explorer. You can now follow the instructions for <em><a href="#install-ubuntu">Ubuntu</a></em>. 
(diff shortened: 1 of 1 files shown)


install.md
```diff
@@ -22,23 +22,11 @@ terminal interface.
 <summary>
 Windows
 </summary>
-We currently only support Windows 10 or newer using <a href="https://learn.microsoft.com/en-us/windows/wsl/install">Windows Subsystem for Linux (WSL)</a>.
+Please install Windows Subsystem for Linux (WSL) following the <a href="https://learn.microsoft.com/en-us/windows/wsl/install">official instructions</a>. You should now have an App "Ubuntu" in your start menu and a folder "Linux" on the left sidebar of your file explorer. You can now follow the instructions for <em><a href="#install-ubuntu">Ubuntu</a></em>. Please note:
```
fingolfin: This is only correct if the user chose to install Ubuntu and not another distribution. So we should inform the user about that. Power user who prefer a different distro should be able to deal with the consequences. 

Perhaps like this:
```suggestion
Please install Windows Subsystem for Linux (WSL) following the <a href="https://learn.microsoft.com/en-us/windows/wsl/install">official instructions</a>.
When asked which Linux distribution to install, choose "Ubuntu".
After the installation finished, you should have an entry "Ubuntu" in your
start menu and a folder "Linux" on the left sidebar of your file explorer.
You can now follow the instructions for <em><a href="#install-ubuntu">Ubuntu</a></em>. Please note: 
```

install.md
```diff
@@ -22,23 +22,11 @@ terminal interface.
...
 <ol>
-<li>Search for "Turn Windows features on or off"</li>
-<li>On the left panel, select "Turn Windows features on or off"</li>
-<li>Select "Windows subsystem for Linux" and press "Ok"</li>
-<li>Click "Restart the PC"</li>
-<li>Click the Windows store icon (shopping bag)</li>
-<li>Search for "Ubuntu" in the store - it's free!</li>
-<li>Select "Ubuntu" and "Get" the app</li>
-<li>Click "Launch" and follow the prompts</li>
+<li> When you start Ubuntu, your terminal will be in the folder "Linux/Ubuntu/home/your_username>". Please avoid changing anything outside that folder unless you are absolutely sure.</li>
```
```suggestion
<li> When you start Ubuntu, your terminal will be in the folder "Linux/Ubuntu/home/<your_username>". Please avoid changing anything outside that folder unless you are absolutely sure.</li>
```
I don't understand this. When I enter `pwd` in the Ubuntu terminal, I see `/home/<username>`. So I think you are referring to the name of this directory when viewed from Windows. We should make that clear.

But isn't this already very specialized? And anyway, would an average user even know how to access that directory (or anything outside it) from Windows? (BTW this can be done by entering `explorer.exe .` in the Ubuntu prompt: that will open the current Linux directory in the Windows explorer.)

install.md
```diff
@@ -22,23 +22,11 @@ terminal interface.
...
+<li> Your browser may be unable to download files directly to the "Linux" folder. Instead, download it to somewhere temporarily and move the file to "Linux" manually.</li>
```
fingolfin: Again, this seems very specialized. I'd rather keep our install instructions as lean as possible, because I've now experienced multiple times that people skip steps and don't read text that is too long.

How about we add a separate page (or a paragraph on the same page, but elsewhere / downwards), with "further tips & tricks for Windows users", were this and more can live, and then just link there?

install.md
YueRen: I don't think it is necessary to inform the user for the following reason:

1. I also think the guide should be as lean as possible.
2. There is no active choice the user has to do. Ubuntu is installed by default when running `wsl --install`.
3. The linked instructions contain information for installing a different system other than Ubuntu. Powerusers who want to do so should be able to follow them without us mentioning it.

install.md
YueRen: I added the information so that the average user knows how to access the directory from Windows (so that they can download script or email their code). But I agree that telling users to simply type `explorer.exe .` is the easier option and we should add that instead of the actual path.

install.md
YueRen: I don't mind omitting that information from that page if you think it's too much.  Whether we need a "further tips & tricks" page we can decide at a later time.

fingolfin (approved):
Nice and concise, great. Thank you very much

PR #252, opened 2023-03-24
Revert to toric tutorial before renaiming

HereAround:
@YueRen , this tutorial should work with the latest official release 0.11.3. If your time permits, I would appreciate if you could confirm.

Once OSCAR v.0.12.0 has been released, this commit should be reverted.


YueRen:
Works like a charm in v0.11.3, thanks for the quick fix!

PR #267, opened 2023-07-31
Include DMG installer for macOS

aaruni96:
Add a new section "Experimental Methods" in install instructions, and include a macOS DMG under it.

Do we have an email someone looks at regularly to replace the `bugs@email.tld` placeholder, or should I link the Oscar.jl issues page there (or remove the "Please report bugs.." altogether) ?


mjrodgers:
I'm not sure I see the need for this; someone using macOS can just install the official Julia .dmg and then as soon as they type `using Oscar` it will install, correct? What exactly does this linked .dmg install?

HereAround:
@aaruni96 What is the status of this PR?

aaruni96:
The PR was in response for an "easier" method to install Oscar on MacOS. It was then abandoned when interest in this died away. I think we can close this as abandoned?

PR #312, opened 2024-04-04
Ak96/remove filters

aaruni96:
Small change to remove the tutorial filter

### DIFF
--- a/_includes/tutorial.html
+++ b/_includes/tutorial.html
@@ -1,7 +1,7 @@
 For each topic, you can decide to open a static version of the jupyter notebook, powered by <a href="https://nbviewer.jupyter.org/">nbviewer</a>.
 Alternatively, you can inspect the jupyter notebook directly on <a href="https://github.com/">github</a>.
 
-<div class="oscar_detail">
+<div class="oscar_detail" style="margin-bottom: 1em;">
   <details>
     <summary>
       How to interact with a "live" version
@@ -54,7 +54,7 @@ Click on one of the links below to filter notebooks (and re-click to disable fil
        >
       <img class="item_example-image-link"
            src="{{ site.baseurl }}/public/thumbnails/{{ component.logo }}">
-      <div> <!-- <div> for centering -->
+      <div>
         {{ component.name }}
       </div>
     </a>
--- a/_includes/tutorial.html
+++ b/_includes/tutorial.html
@@ -35,6 +35,7 @@ Alternatively, you can inspect the jupyter notebook directly on <a href="https:/
   </details>
 </div>
 
+<!--
 <br/>
 Click on one of the links below to filter notebooks (and re-click to disable filtering).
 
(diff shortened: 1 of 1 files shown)


HereAround:
I just eyeballed the changes. As you say, the filters are gone. That is great. Now it looks as follows:
NewLooks

Please rebase to the latest gh pages. (I added 3 more tutorials today, that are apparently not included in your PR.)

Suggestion 1:
I believe, we want to improve this a bit more. This can either be done in this PR or a separate PR. Here is what I have in mind.

My recollection is that the suggestion in the Wednesday meeting was to remove the logos (and so also the filters) for the following reasons:
* All tutorials are OSCAR tutorials. So no need to display the OSCAR logo.
* Likewise, since we want to promote OSCAR tutorials, rather than corner stone tutorials, no cornerstone logos neither. (e.g. Polymake, in the above screenshot).

If that makes sense to you, I would thus suggest to remove the OSCAR and Polymake logos in the above screenshot. Possibly including removing the .png files from the website.

There is a chance, that we want to bring back a visual indication about the content of the tutorial (e.g. a logo for number theory, one for polyhedral geometry, one for tropical geometry, one for group theory, one for toric geometry, one for schemes etc.) So I suggest not to completely remove everything, but just to the extend that it is currently not being displayed and could be brought back.

Suggestion 2:
If you visit the latest website (https://www.oscar-system.org/tutorials/), then you can see that the tutorial on "Analyzing Rubik's Cube" almost fills up the entire vertical space that is reserved for one tutorial. Of course, we could add extra vertical space, but I feel that this would only be a temporary fix. Maybe it is more efficient to just allow the tutorials to fill the entire page width? I.e. one tutorial per "line", rather than two tutorials side by side.

What do you think?

HereAround:
The tutorial on "Generation of sporadic simple groups by 𝜋- and 𝜋'-subgroups" uses up even more vertical space.

_includes/tutorial.html
```diff
@@ -53,15 +54,15 @@
        >
       <img class="item_example-image-link"
            src="{{ site.baseurl }}/public/thumbnails/{{ component.logo }}">
-      <div> <!-- <div> for centering -->
+      <div>
```
HereAround, line 57: Why is this change needed?

HereAround: Just removing an old comment, that is no longer needed I guess?

HereAround:
On 2nd thought: I think we can postpone the more cosmetic changes to a later PR, in particular given that the dynamic status PR is likely a lot more important/harder...

So I have just rebased...

HereAround:
And will now merge this one. Let us return to the cosmetic changes mentioned above later.

_includes/tutorial.html
aaruni96: More than just not needed. You can't nest comments (guess how I found out).

```
<!-- start of outer comment
        <!-- start of inner comment
        end of inner comment -->
end of outer comment -->
```

The entire comment ends at `end of inner comment-->`, and the rest is parsed on a "best effort basis" to render a mess.

HereAround: I see. Thx for explaining.

PR #313, opened 2024-04-04
Dynamic Status for Tutorials

aaruni96:
A mildly hacky method to split tutorials into two sections: current and out of date, depending on whether the CI run for it passed or failed .


HereAround:
First comment (checking this PR out locally): Rebase to the current master leads to a merge conflict in `_data/examples.yml`. I am surprised this is not picked up by github...

aaruni96:
I suspect it is because 8d739e1 is a partial revert of e6c88c3 . The earlier commit added a "status" badge to the example.yml, just to test things out, and the later commit reverted the change to example.yml after I figured out a nicer way to do things. So in the PR as a whole, there's no changes to the file, but its a "different version" according to git?

HereAround:
A rebase of this PR is definitely necessary. From https://github.com/oscar-system/TutorialTesterforOscar/actions/runs/8548274715/job/23421742415, I see that one tutorial failed. That is great news, because it would allow to eyeball how a failed tutorial will look like. Right now, only "passed" tutorials are being displayed when I try out this branch locally.

HereAround:
> I suspect it is because [8d739e1](https://github.com/oscar-system/oscar-website/commit/8d739e165af2ab2e77a9b63a4e4d7676943ebd3d) is a partial revert of [e6c88c3](https://github.com/oscar-system/oscar-website/commit/e6c88c3a980c338789d2fb8eee619daa3f65a3bc) . The earlier commit added a "status" badge to the example.yml, just to test things out, and the later commit reverted the change to example.yml after I figured out a nicer way to do things. So in the PR as a whole, there's no changes to the file, but its a "different version" according to git?

Hhmm... Maybe? But that is indeed somewhat weird.

aaruni96:
I have rebased to the latest gh-pages

HereAround:
Thank you! Here is how this looks for me locally:

FailedTutorialRubric

I was hoping to check out the section of failed tutorials, but no tutorials seems to be listed there. Based on https://github.com/oscar-system/TutorialTesterforOscar/actions/runs/8548274715/job/23421742415, should there not be exactly one tutorial listed as currently failing?

HereAround:
Aside: I like the wording choice "Out of date". That is much better than "failing tutorials" or alike.

aaruni96:
I should probably write some better documentation. The `Makefile` uses `jq` to parse the json object returned by the Github API. If the file created by make does not exist, the "default" state is to assume a tutorial is passing.

This is how it looks with the `examples_status.yml` being generated correctly by the makefile.

image

HereAround:
Ah. So one has to invoke
```
bundle exec make
bundle exec jekyll serve
```
to check out those changes locally.

Looks good to me (modulo the cosmetics mentioned on the previous PR).

~~One more question: This display already updates dynamically to the latest test results?~~ Cool!

HereAround:
@aaruni96 From what I see, the only point remaining is to trigger an update to the OSCAR website whenever the tutorial tester completed its run. Is that correct?

aaruni96:
This probably needs a look over from @fingolfin and then needs changes in the deployed environment and CI scripts based on his input. I would not merge this until after Max has seen this and is happy with this.

_sass/oscar.scss
```diff
@@ -158,5 +158,32 @@ details[open] summary::after {
...
+
+div.borked > div.gallery_example_cards > div.item_example_card {
+  color: white;
+  margin: 15px;
+  background: #979797;
+}
+
+div.borked > div.gallery_example_cards > div.item_example_card > div.item_example_right > div> a, div.borked > div.gallery_example_cards > div.item_example_card > div.item_example_right > a{
+  color: skyblue;
+}
```
fingolfin: add trailing newline?

_includes/tutorial.html
```diff
@@ -96,3 +97,35 @@
...
+      <div><a href="https://nbviewer.jupyter.org/github/{{ post.repository }}/blob/{{ post.branch }}/{{ post.filename }}.ipynb">nbviewer</a> </div>
+      <a href="https://github.com/{{ post.repository }}/blob/master/{{ post.filename }}.ipynb">Github</a>
+
+    </div>
+  </div>
+  {% endfor %}
+</div>
+</div>
+</details>
+<span class="date">Last updated on {{ site.time | date: '%d-%m-%Y' }}</span>
```
fingolfin: add trailing newline?

Makefile
```diff
@@ -0,0 +1,20 @@
...
+  $(error This Make does not support .RECIPEPREFIX. Please use GNU Make 4.0 or later)
+endif
+.RECIPEPREFIX = >
+#/preamble
+
+.PHONY: build
+build:
+>myurl=$$(curl -Ssl "https://api.github.com/repos/oscar-system/TutorialTesterforOscar/actions/workflows/CI.yml/runs?per_page=1" | jq  '.workflow_runs[0].jobs_url' | sed 's/"//g')
+#>echo $$myurl
+>curl -SsL $$myurl | jq '.["jobs"][1:] | .[] | .name+":"+.conclusion' | sed 's/^.*\s.*\s.*\s//' | sed 's/):/: /' | sed s'/"$$//' > _data/examples_status.yml
```
fingolfin: It may well be that the GitHub fails to return a status (due to a temporary network issue, a temporary issue on their server, etc. ). Maybe tell `curl` to retry, it has a bunch of feature for that

```
     --retry <num> Retry request if transient problems occur
     --retry-all-errors Retry all errors (use with --retry)
     --retry-connrefused Retry on connection refused (use with --retry)
     --retry-delay <seconds> Wait time between retries
     --retry-max-time <seconds> Retry only within this period
```

Of course it still may fail even with retries. One could think about how to deal with that at some point, but it hopefully doesn't need to be a top priority right now.

Makefile
```diff
@@ -0,0 +1,20 @@
...
+>jekyll build
```
fingolfin: Don't run `jekyll` here, we need to invoke it via bundler from `etc/update.sh`

Makefile
```diff
@@ -0,0 +1,20 @@
+# Sane Makefile Preamble from https://tech.davis-hansson.com/p/make/
```
fingolfin: Why is this whole thing even a Makefile instead of just a shell/Python/whatever script? There seems to be nothing that make adds here other than a lot of effort to force it to act as close as possible to a bash script?

Note that we need to invoke this from `etc/update.sh`. So how about turning this into, say, a script `etc/fetch_examples_status.sh` which then is invoked from `update.sh` ? Or even directly make it a part of `update.sh`.

aaruni96: I forgot about the `update.sh` script, and a Makefile is my default answer for "we need to do this preprocessing step before we invoke build" so whoever is using the project just needs to run `make`.

I have removed the makefile and moved things to the update script.

aaruni96:
@fingolfin should I also add `_data/examples_status.yml` to `.gitignore` ? 

Argument for yes : its a script generated file and should not be checked into get.
Argument for no : its a single file which is automatically generated and nobody cares.

fingolfin:
Can I build the website on my computer without a `_data/examples_status.yml` present? I suspect there will be an error?

As such perhaps a dummy version should be in the repository. That one could even have some useful default data, like "status: unknown", "last_changed: unknown" so that when there are network issues, this can be used as a fallback?

aaruni96:
Jekyll continues silently without error if it can't find the data. I've made the css rules such that when that happens, it defaults to displaying all the tutorials as "working" (as happened with @HereAround earlier in this thread when he ran jekyll serve without first fetching the data).

fingolfin:
Thanks, all makes sense!

PR #324, opened 2024-04-24
Make tutorial cards of dynamic height

aaruni96:
cc @HereAround


HereAround (approved):
Thank you @aaruni96 , looks good to me!

PR #326, opened 2024-05-06
Update mentioned julia version in install.md

thofma:


install.md
```diff
@@ -89,7 +89,7 @@ Suitable compilers include
 ## Step 2: Install Julia
 
 OSCAR requires at least [Julia](https://julialang.org) 1.6.0, but we recommend running it with the latest stable Julia release,
-which is 1.9.3 at the time this is written.
+which is 1.10.3 at the time this is written.
```
fingolfin: Shouldn't we rather drop the version and just end with "... the latest stable Julia release." to avoid having to update this constantly?

thofma: Yes, good idea, then we only have to adjust the LTS version in the future.

fingolfin (approved):

PR #337, opened 2024-05-22
Decode json as array, not object

aaruni96:
This will enable processing webhooks from the tutorial tester


HereAround:
Thank you @aaruni96 . Looks good to me, but let @fingolfin double check.

fingolfin (approved):
No idea if this is correct, let's find out

PR #340, opened 2024-05-27
Fix Tutorial filter for responsiveness

### LINKED ISSUE #338: Tutorials look very condensed on phone in portrait mode

aaruni96:
Fixes the tutorial filter to behave better in narrow screens: make the words break, and equalize the heights.

I have replaced the HTML table with a flexbox div, and transferred the styles over (the table would either keep the same heights, or break the text, but not both for whatever reason).

Here's a "side by side" of flexbox (top) v/s table (bottom).

image

Fixes #338 , hopefull.

cc @HereAround

### DIFF
--- a/_data/examples_components.yml
+++ b/_data/examples_components.yml
@@ -1,12 +1,12 @@
 - logo: FTheoryTools.png
-  name: FTheoryTools
+  name: FTheory Tools
 - logo: GroupTheory.png
-  name: GroupTheory
+  name: Group Theory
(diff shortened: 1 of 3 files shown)


aaruni96:
In mobile view, the tutorials page now looks like this, which I think looks fine ?

image

HereAround:
Great. Thank you @aaruni96 . I like this!

HereAround (approved):

Issue #346, opened 2024-06-19
Broken links and other observations


HereAround:
I use this issue to collect my observations regarding the website. This may need to be restructured eventually.

All links on https://www.oscar-system.org/meetings/ broken (but the one pointing to the SFB-meetings at the very bottom). E.g. https://www.oscar-system.org/https://www.oscar-system.org/meetings/2023-09/ when I click on "Meeting 09/2023".

HereAround:
cc @aaruni96

HereAround:
I just found another strange link (I am going through the entire website now):

https://www.oscar-system.org/news/

Then click on "OSCAR news atom feed". Is this desired?

aaruni96:
Yeah, I think that one is working as intended. Its the raw feed for people to subscribe to updates via atom ( https://en.wikipedia.org/wiki/Atom_(web_standard) )

HereAround:
Maybe the link to the `AbstractSolving` documentation (as listed here https://www.oscar-system.org/documentation/) should point to https://algebraic-solving.github.io/?

HereAround:
Also, the `homalg_project` is not emcompassed by OSCAR. Remove the links at
* https://www.oscar-system.org/documentation/
* https://www.oscar-system.org/community/

I wonder if also OSCAR.jl should be removed from the list at the bottom of https://www.oscar-system.org/documentation/. After all, the links to the OSCAR documentation are at the top of that page (and highlighted so).

HereAround:
https://www.oscar-system.org/community/ I find the OSCAR development email as oscar-dev@mathematik.uni-kl.de. Maybe this should be updated to an rptu email eventually?

HereAround:
No talks in 2024? (cf. https://www.oscar-system.org/talks/)

HereAround:
Link to cdd at https://www.oscar-system.org/credits/ is broken. Same for factory, gfanlib and mpir.

HereAround:
I have began to fix at least some of these points here: https://github.com/oscar-system/oscar-website/pull/347

HereAround:
Closed by #347 and #348.

PR #348, opened 2024-06-19
baseurl is prepended to path when running the relative_url filter

aaruni96:
Should fix the broken links in meeting pages mentioned in #346 

cc @HereAround


HereAround (approved):
Thank you @aaruni96

PR #360, opened 2024-06-28
Update .gitignore

HereAround:
@aaruni96 what do you think? Was there deep reasons to not include those paths that I include? Also, I am surprised you did not add the path `_data/` previously. I was under the impression that this is necessary, not?


aaruni96:
No deep reason, I asked if its a good idea to add it to the gitignore, and then it got lost among other comments (and I forgot about it)

https://github.com/oscar-system/oscar-website/pull/313#issuecomment-2039330033

.gitignore
```diff
@@ -10,4 +10,6 @@
 /Gemfile.lock
 
 # Files generated by update script
-examples_with_updated_last_modified.yml
+_data/examples_with_updated_last_modified.yml
+_data/examples_status.yml
+_data/examples_with_updated_dates.yml
```
aaruni96: I don't think we use `examples_with_updated_dates.yml`, it was the old name of the "last_modified" yaml, if memory serves ?

HereAround: Ok. Then I will just merge this now.

HereAround: (Removed `examples_with_updated_dates.yml`.)

PR #367, opened 2024-08-01
Minor fix for schemes tutorial

HereAround:


HereAround:
I believe this should have fixed the nbviewer issue. (There is also caching involved. So it maybe require a restart of the browser, clearing cache, or just waiting a little...)

PR #370, opened 2024-08-07
Update contributer list for Q2 2024

aaruni96:
cc @HereAround


HereAround (approved):
Great! Thank you very much @aaruni96

PR #378, opened 2024-10-29
Remove month from book bibliograhpy

aaruni96:
Remove month from bibliography, as it is no longer accurate. See https://github.com/oscar-system/Oscar.jl/pull/4252 .


HereAround:
Thank you @aaruni96 !

PR #373, opened 2024-10-04
Ak96/automate contributer list

aaruni96:
This script will eventually be used to automate the updating of the contributer's list on the website.

cc @HereAround

PR #387, opened 2024-11-06
Move contributors list to _data

aaruni96:
A split from #373, to make it smaller / reduce change of merge conflicts while that is prepared.

We can rebase #373 to gh-pages once this is merged

cc @HereAround


HereAround:
Looks great. Thank you @aaruni96

PR #393, opened 2024-11-27
Ak96/ci contrib list

aaruni96:
Changes in this PR enable a bot to make PRs like #392 . This can be merged once we figure out what to do with the "continue case", i.e., if a commit was made by someone in git, but pushed by someone else to github, the github API returns the author of the commit as `None`. Currently we skip cases like this, but this should be handled. ( cf. `etc/update-contributors.py:75` | currently happens with `['Jens Brandt', 'jens.brandt@rwth-aachen.de]` ), and he is not included in the contributor's list (mostly so we can reproduce the problem).

cc @HereAround

### DIFF
--- a/.github/workflows/update-contrib.yml
+++ b/.github/workflows/update-contrib.yml
@@ -42,18 +42,11 @@ jobs:
         git add _data/people_list.yml
         git commit -m "Update contributors on $(date +'%Y-%m-%d')" || ${echo "No changes to commit" && false}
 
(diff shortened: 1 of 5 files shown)


aaruni96:
Currently, things appear to ~~mostly~~ work. ~~We just cant figure out how to make a multiline PR message in #392 . But otherwise,~~ This appears ready for review, and then can be squashed-merged.

Please have a look and let us know what you think @fingolfin

PR #397, opened 2024-12-04
CI: limit concurrency for update-contrib workflow

fingolfin:
If multiple PRs are merged into master in a row, we don't want them all to run, just the one on the latest commit.

### DIFF
--- a/.github/workflows/update-contrib.yml
+++ b/.github/workflows/update-contrib.yml
@@ -9,6 +9,12 @@ on:
   schedule:
     - cron: '0 0 1 * *'  # At 00:00 UTC on the 1st of every month
 
+# only run at most one instances of this workflow at a time for each branch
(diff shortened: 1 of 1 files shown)


aaruni96:
Great idea, but I will have to dive into the documentation for concurrency to actually review the change.

Just to confirm, since the action only runs when a someone pushes to gh-pages (so, after a regular PR is merged), and if multiple PRs are merged in a short succession, we cancel the ones still in progress, and only keep the run on the latest commit on gh-pages. This is what your change will do, right ?

fingolfin:
Yes. We use something similar in Oscar.jl's `CI.yml`. The code in this PR is a bit different though and copied from https://github.com/gap-system/PackageDistro/blob/main/.github/workflows/assemble-distro.yml (where it has been in extensive use for quite some time)

aaruni96 (approved):
Looks good to me

Issue #220, opened 2021-10-11
Better install instructions for Windows users


fingolfin:
1. say more precisely which Julia binary to download ("look for Linux 64bit glibc"), or even better, provide a direct download link
2. say how to actually download the result inside WSL
3. say how/where to extract it so that the user can access it from anywhere ("add it to PATH")
4. alternatively to the above: suggest a "one click install method" if there is one (probably not...) or at least one that is simpler; e.g. it might be easier to tell people something like `sudo apt-get install python3 && pip3 install jill --user -U && jill install` or so (to be tested!!!!)
5. It would be great to have a way to add an icon to the desktop or start menu which users can select to start a Julia session (inside WSL), or even an Oscar session

fingolfin:
I guess this partially duplicates issue #150 but I don't have time to properly join them right now

fingolfin:
Resolved by PR #399 and #404

PR #399, opened 2024-12-11
Improve installation instruction for Windows users

### LINKED ISSUE #4329: Additional setup instructions for Windows/WSL
When trying to install Oscar on Windows, the following problem occurs quite frequently: 

Oscar needs 6GB of RAM. The standard WSL setup has 4GB of RAM. It is not immediately obvious that this is the issue. Instead of throwing an error, Oscar just gets stuck compiling.

There is an easy fix to this: one needs to change the corresponding config file. 

I propose to change the instructions for the Windows setup in the following (or a very similar) way:

- In WSL, check `free`
- If the free RAM is smaller than 6GB, save the attached config file [wslconfig.txt](https://github.com/user-attachments/files/17817207/wslconfig.txt) in your C:User/"your username" branch and rename it to `.wslconfig` (with a dot in the beginning and without any file type ending! This is possible, for instance, using Notepad, changing the data type to "All Types")
- Close WSL, open the PowerShell
- type wsl --shutdown
- restart WSL and continue with the Oscar installation

worked out with @MichaelSchloesser

HereAround:
This aims to close https://github.com/oscar-system/Oscar.jl/issues/4329. It may also cover/advance https://github.com/oscar-system/oscar-website/issues/220.

### DIFF
--- a/install.md
+++ b/install.md
@@ -16,18 +16,32 @@ or by sending an email to <mailto:oscar@oscar-system.org>.
 The following instructions assume that you are at least somewhat familiar with using a
 terminal interface.
 
-The installation steps (specifically the precompliation) requires at least 6GB of free memory.
-
+<div class="message">
+  <strong>WARNING:</strong>
+  The installation process, particularly the precompilation step, requires 
+  <strong>at least 6GB of free memory</strong>.
+<br>
+<br>
+  For <strong>Windows</strong> systems, it is necessary to use the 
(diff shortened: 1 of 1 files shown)


HereAround:
I merge this PR now. @aaruni96, maybe you can bring this up in triage next week (the corresponding OSCAR issue linked above has a triage label) to verify if people are happy with these changes/what more is needed.

Ideally, both of the issues linked above could be closed.

PR #401, opened 2024-12-11
Fix the false error we currently get

aaruni96:
This should fix the update-contributors action showing an error if there is nothing to commit.


aaruni96:
This works, as can be seen here : https://github.com/oscar-system/oscar-website/actions/runs/12279739781

PR #411, opened 2025-01-21
Update and add new credits to Oscar

StevellM:
One of the paper has been published since June 2024, another paper has had a major update and two other papers were not added to the list.

### DIFF
--- a/credits.md
+++ b/credits.md
@@ -2,25 +2,37 @@
 layout: page
 title: Credits
 citations:
+    - name: On birational automorphisms of double EPW-cubes
+      url: https://arxiv.org/abs/2405.15510
+      authors: Simone Billi, Stevell Muller, Tomasz Wawak
+      month: Dec.
+      year: 2024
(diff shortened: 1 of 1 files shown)


HereAround:
@StevellM looks good. Thank you!

One question: On closer inspection, the paper with @simonbrandhorst follows the citation suggestion for OSCAR (cf. https://www.oscar-system.org/, bottom of the page). In the other two papers, you dropped the citation of the OSCAR book. Just for my curiosity - why?

HereAround (approved):

StevellM:
@HereAround I guess at the time of the first versions of the papers, I forgot to cite the Oscar book (some of the papers where even written before we had the citation), and I have never thought to update it... For the non-published ones, I will update the citation for the book instead of the github repo, in future revisions.
(I actually forgot to update another credit, but I will wait for the paper to be published before updating. I will take care of updating the Oscar citation in that paper also.)

PR #414, opened 2025-01-22
Improve instructions for system wide installation

fingolfin:
This is an unfinished draft. I started this when I was in Braunschweig and tried to help them set up a system wide OSCAR installation.

I plan to finish this at another point but don't want to forget about it, so this PR is mostly there as a reminder. Not much point in reviewing it in its current state.

### DIFF
--- a/install-generic.md
+++ b/install-generic.md
@@ -55,68 +55,79 @@ and then run the script in a terminal (with administrator rights).
 {% highlight bash %}
 #!/bin/bash
 
+set -e
+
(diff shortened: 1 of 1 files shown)


fingolfin:
Ready from my POV (this is what we deploy in Kaiserslautern). CC @ederc

ederc (approved):
Works for me.

HereAround (approved):
LGTM. Thank you!

Issue #422, opened 2025-02-14
URLs changed, many broken links


fingolfin:
Unfortunately with the refactoring by @aaruni96 and @HereAround many URLs changed, and as a result external links are broken.

E.g. luckily I did not yet send out the email inviting to the Summer School but the link we have on the SFB homepage to it still is broken https://www.oscar-system.org/meetings/2025-09/

It now is <https://www.oscar-system.org/events/meetings/2025-09/>

We really should set up redirects for as many of these as possible.

fingolfin:
To do that you can just edit `.htaccess`.

aaruni96:
Ah, I was planning to use the `<meta>` tag for this

https://stackoverflow.com/questions/5411538/how-to-redirect-one-html-page-to-another-on-load#5411567

aaruni96:
But your suggestion might be much better

HereAround:
I believe most if not all of the links have been fixed. @fingolfin @aaruni96, can this be closed?

aaruni96:
I think this is done (we made the redirects, and haven't heard any more complaints)

PR #423, opened 2025-02-17
Fix broken link

StevellM:
I guess was meant to redirect to the "bibliography" page. Currently URL is not found.


fingolfin:
Thanks @StevellM .

Much more is broken, see issue #422 and we need to fix those all systematically, ASAP. And not just the internal ones, but also external ones. 

CC @HereAround @aaruni96

PR #424, opened 2025-02-17
Redirect old meeting URLs

fingolfin:
First step towards resolving issue #422, as a blue print for how to continue from there.

CC @HereAround @aaruni96 

I suggest we merge this ASAP and check if it works


HereAround:
Thank you @fingolfin - am looking into it now.

PR #427, opened 2025-02-17
change page link to match sidebar link

aaruni96:
Minor typo which lead to a "not found" page. If we want the URL to actually be `oscar-events`, then we must change the link on the sidebar, and the actual target for the page.

### DIFF
--- a/events/index.md
+++ b/events/index.md
@@ -3,7 +3,7 @@ layout: page
 title: Events
 ---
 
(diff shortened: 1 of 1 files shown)


fingolfin (approved):
Looks good to me

HereAround:
Thank you @aaruni96 !

HereAround (approved):

PR #428, opened 2025-02-17
Try Redirect insteadof RedirectMatch

aaruni96:
As suggested by @fingolfin  at https://github.com/oscar-system/oscar-website/pull/426#discussion_r1958097328


aaruni96:
Wait, this change may not be required. 

The website works as intended without this change on chromium, but not on firefox. It might just be a caching issue on my device.

HereAround:
Just briefly discussed on this with @aaruni96 . We decided to not merge this PR for now. On Wednesday, we shall reevaluate the situation.

(For all it matters: We could sufficiently convince ourselves that the redirects are working as intended. Also on Firefox. Hence, what @aaruni96 reported initially, must be a local caching problem.)

aaruni96:
I think `RedirectMatch` works well enough, and I don't fully understand .htaccess rules, so swapping out `RedirectMatch` with `Redirect` may either

- cause no changes
- break something

I think its therefore best to close this PR as unmerged, until someone has a better reason to revisit this idea.

PR #429, opened 2025-02-19
Add more OSCAR references

HereAround:
cc @aaruni96 . Let us check that the link to zb math works as intended.


HereAround:
The querry on zb-math leads to the papers citing OSCAR because of https://zbmath.org/software/36845, in case anyone is curious. This is what I wanted to verify.

With that being said, I think this can be merged.

PR #434, opened 2025-02-21
Participating in Coding Sprint 03/25

ederc:


fingolfin:
Excellent that you can come!

@aaruni96 @HereAround can you please set up a separate participants page (as in other past meetings)?

aaruni96:
Added participants page, with @ederc  in it.

PR #447, opened 2025-03-05
Wording changes and general improvements

HereAround:
With this PR, we still have to manually update the information in `data/oscar_version.yml`. Still, this is a lot easier than having to modify every single version number individually on the website. Plus, I image it will not be too hard to add a script which - for instance on every merge to gh-pages - updates the content of this file to the latest stable release.

What are your thoughts @aaruni96 ?

PR #449, opened 2025-03-06
Show release version

HereAround:
Based on https://github.com/oscar-system/oscar-website/pull/447, and aims to show the latest stable release version of OSCAR in various places across the OSCAR website.

I mark this as draft, since we should discuss a script to update the version information dynamically. Maybe in the form of a cronjob, or running this script with every merge to the gh-pages branch.

cc @aaruni96 @fingolfin


aaruni96:
Adding version information in some cases makes sense ( index.md, _includes/tutorial.html, credits/Citing-OSCAR.md, getting-started/install-*.md), but in other cases (all tutorial pages) maybe not ? We already tell in a few places what the latest OSCAR is, and hide outdated tutorials.

aaruni96:
Noting here for future work on the script : the latest release is always available at https://api.github.com/repos/oscar-system/Oscar.jl/releases/latest .

HereAround:
Let me improve this. I noted that I left some references of the OSCAR version in older commits, so have to overhaul the history to properly reflect this. Will then also remove the version number from the titles on the tutorial pages.

HereAround:
Should be updated now.

fingolfin:
Without having looked at the changes yet, let me just say that in general I would minimize references to the OSCAR version to a few places (maybe the front page to tell people which version is current, and the installation page(s)... and that might be it?). This is orthogonal to defining it in a single place (which I understand this PR does, which is a good thing of course).

For updating it, for the GAP website we have a GitHub workflow in place which runs daily and checks for a new release, and if there is one, opens a PR to update the version. That workflow can also be triggered manually for the impatient. Perhaps we can do the same?

HereAround:
> Without having looked at the changes yet, let me just say that in general I would minimize references to the OSCAR version to a few places (maybe the front page to tell people which version is current, and the installation page(s)... and that might be it?). This is orthogonal to defining it in a single place (which I understand this PR does, which is a good thing of course).
> 
> For updating it, for the GAP website we have a GitHub workflow in place which runs daily and checks for a new release, and if there is one, opens a PR to update the version. That workflow can also be triggered manually for the impatient. Perhaps we can do the same?

Agreed.

To state two more natural places for a reference to the OSCAR version: On the documentation page we can say "click here for the manual to OSCAR version ...". Similarly, I found it reasonable to update the bibtex entries on the "how to cite OSCAR" page, so that they always reflect the latest stable release of OSCAR.

Still, it might be better to first look into https://github.com/oscar-system/oscar-website/pull/447. Once merged, the changes in this PR will become fairly minimal and easy to discuss (cf. https://github.com/oscar-system/oscar-website/pull/449/commits/d2749ead6c357b5dbd7872eda4802c63409c0ab6), so that we can see exactly where we want to refer to the latest OSCAR release.

HereAround:
@aaruni96 @fingolfin I just rebased this, and the changes in this PR should now be much easier to eyeball/discuss.

credits/Citing-OSCAR.md
```diff
@@ -3,12 +3,12 @@ layout: page
...
 
-If you have used **OSCAR** in your research, please cite it using the following format:
+If you have used **OSCAR v{{ site.data.oscar_version.latest_release }}** in your research, please cite it using the following format:
 
 ```md
 [OSCAR]
-    OSCAR -- Open Source Computer Algebra Research system, Version 1.0.0,
-    The OSCAR Team, 2024. (https://www.oscar-system.org)
+    OSCAR -- Open Source Computer Algebra Research system, Version {{ site.data.oscar_version.latest_release }},
+    The OSCAR Team, 2025. (https://www.oscar-system.org)
```
HereAround: Should 2025 be replaced by a reference to the current year?

credits/Citing-OSCAR.md
```diff
@@ -23,8 +23,8 @@ If you are using **BibTeX**, you can use the following BibTeX entries:
   key          = {OSCAR},
   organization = {The OSCAR Team},
   title        = {OSCAR -- Open Source Computer Algebra Research system,
-                  Version 1.0.0},
-  year         = {2024},
+                  Version {{ site.data.oscar_version.latest_release }}},
+  year         = {2025},
```
HereAround: Likewise here, replace 2025 by the current year?

fingolfin: No, but it should be taken from the YAML.

This is also how we do it for the GAP website, see https://github.com/gap-system/GapWWW/blob/master/_data/release.json

HereAround: Just updated.

HereAround:
Preview at: https://herearound.github.io/oscar-website/

Please take a look at the bibtex entry at https://herearound.github.io/oscar-website/credits/Citing-OSCAR/. There is an additional white space in the field "year", which I cannot seem to remove, e.g. by putting the expression in quotes (in this case the quotes are displayed on the website) or removing the corresponding white space in the .md file (leads to a liquid error). @aaruni96 maybe you have an idea?

credits/Citing-OSCAR.md
```diff
@@ -22,11 +22,10 @@ If you are using **BibTeX**, you can use the following BibTeX entries:
 @misc{OSCAR,
   key          = {OSCAR},
   organization = {The OSCAR Team},
-  title        = {OSCAR -- Open Source Computer Algebra Research system,
-                  Version 1.0.0},
-  year         = {2024},
+  title        = {OSCAR -- Open Source Computer Algebra Research system, Version {{ site.data.oscar_version.latest_release }}},
+  year         = { {{ site.data.oscar_version.latest_release_date | split: " " | last }}},
```
fingolfin: Try this
```suggestion
  year         = { {{- site.data.oscar_version.latest_release_date | split: " " | last }}},
```
For background, see https://shopify.github.io/liquid/basics/whitespace/

_data/oscar_version.yml
```diff
@@ -0,0 +1,2 @@
+latest_release: "1.3.0"
+latest_release_date: "28 February 2025"
```
fingolfin: I don't want to bikeshed, but if you only ever need the year, why not restrict to that? Or even if you sometimes need the year and sometimes the full date, perhaps just store both in this file (the plan is to update this file by a script anyway, right? so redundancies won't matter)

Also if you rename the file to e.g. `release` (or `latest_release`?) and the keys in here to `version`, `date`, `year`, then you can replace `site.data.oscar_version.latest_release` by `site.data.release.version`, and also use ``site.data.release.date` and ``site.data.release.year` which IMHO is nicer (less verbose and still more on the point).

But again, this is bikeshedding, I am also fine with the current format 🤷 😄

credits/Citing-OSCAR.md
fingolfin: Seems to have worked :-)

fingolfin (approved):
Looks good to me

credits/Citing-OSCAR.md
HereAround: Indeed. This works. Thx!

_data/oscar_version.yml
HereAround: Great suggestion. Executed.

HereAround:
> Looks good to me

Thank you. Will merge this now. Script for automatically updating this version will follow in separate PR.

Issue #453, opened 2025-03-10
Automatically find latest OSCAR release


HereAround:
Follow-up to https://github.com/oscar-system/oscar-website/pull/449: Update the release information automatically.

HereAround:
Starting points:

@aaruni96 pointed out that the latest version is available here: https://api.github.com/repos/oscar-system/Oscar.jl/releases/latest

@fingolfin pointed out that:

> For updating it, for the GAP website we have a GitHub workflow in place which runs daily and checks for a new release, and if there is one, opens a PR to update the version. That workflow can also be triggered manually for the impatient. Perhaps we can do the same?

HereAround:
Can be closed, as https://github.com/oscar-system/oscar-website/pull/458 was just merged.

Issue #435, opened 2025-02-23
Update OSCAR version in sample citing


albinahlback:
https://www.oscar-system.org/credits/Citing-OSCAR/ reads v1.0.0 as version in the citation block, and I suppose it would be more appropriate to automate the updating this version number (currently the version is 1.2.2).  For reference, we have a similar issue in FLINT (see https://github.com/flintlib/flint/issues/2219) which is currently unresolved.

HereAround:
Good point. We are looking into it.

albinahlback:
We solved this in FLINT by changing our release CI to push changes to our website through SSH.  However, I recall you can make a PR through Github actions, perhaps this would be suitable, but I do not know how to do it (I suppose I can look into it during the workshop).

lgoettgens:
I guess this is already resolved by https://github.com/oscar-system/oscar-website/pull/449, isn't it?

HereAround:
> I guess this is already resolved by [#449](https://github.com/oscar-system/oscar-website/pull/449), isn't it?

Yes, a version is shown on our website by now. However, this version is not yet updated automatically without human interference. I want this, so that a PR like https://github.com/oscar-system/oscar-website/pull/457 will not be necessary in the future. I believe @albinahlback intention is similar, so I suggest to leave this issue open for now.

We have set up a job which opens a PR whenever the OSCAR contributors need updating. Similarly, we plan to have a bot generated PR whenever the OSCAR version needs updating. Ideas for this are in https://github.com/oscar-system/oscar-website/issues/453. @aaruni96 and I will work on this in the not too distant future, unless someone beats us to it. ;) (If fortunate, this may mean this coming Wednesday.)

HereAround:
With https://github.com/oscar-system/oscar-website/pull/458 merged, I think this issue can be closed.

PR #458, opened 2025-03-17
add release info to update script

aaruni96:
Addresses #453 , and also #435 

This will automatically update version with the latest release at runtime, similar to how we update tutorial status.

### DIFF
--- a/etc/update.sh
+++ b/etc/update.sh
@@ -21,5 +21,17 @@ curl -SsL --retry 5 $statusurl | jq '.["jobs"][1:] | .[] | .name+":"+.conclusion
 cd etc
 python3 update-dates.py
 cd ..
+# update version info
+curl -SsL https://api.github.com/repos/oscar-system/Oscar.jl/releases/latest > latest.json
+version=$(jq '.["name"]' latest.json | sed 's/v//')
+date=$(jq '.["created_at"]' latest.json | cut -c2-11)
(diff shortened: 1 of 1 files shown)


HereAround (commented):
Thank you @aaruni96 . Generally speaking looks good to me. But of course, the best way to try it out is to merge it and see it in action.

Before that, and so I understand your change better, two question towards how you modify `_data/release.yml`.
* I imagine that we need to clear the content of that file before we amend more new content, but cannot see this in your changes. Am I overlooking it? Is my understanding wrong?
* The file `_data/release.yml` is checked into the git history. Consequently, we will need a PR to change this file. Or am I mistaken? I imagined a bot-generated PR similar to the one that updates the OSCAR contributors. Maybe this is done implicitly and I am overlooking this?

aaruni96:
> I imagine that we need to clear the content of that file

In line 31, we redirect output using a single `>`, which is to overwrite a file. In the later lines, we use `>>` to redirect, which appends to a file.

> The file _data/release.yml is checked into the git history

This script modifies the file directly on the server at the time of building the website (after all the git activity). We can in fact remove this file from git tracking if we want (or keep the current version so its still available for local testing). This is exactly how we update the tutorial status as well, directly on the website server, independent from git activity.

HereAround:
Thank you for explaining @aaruni96 . Sounds good. Let us try it out.

HereAround (approved):

PR #462, opened 2025-03-25
Events: ensure internal links use baseurl

fingolfin:
... so that testing them before deployment works as expected


HereAround (approved):
Looks good to me. Thank you!

PR #463, opened 2025-03-25
Update contributors list and its generation

fingolfin:
- also scan AlgebraicSolving.jl for contributors
- no special case for Janko anymore
- clarify what kind of contributions we do and don't show
- show number of active/retired contributors


aaruni96 (approved):
Looks good, thanks!

PR #472, opened 2025-03-26
Events: ensure internal links use baseurl

lgoettgens:
The same as https://github.com/oscar-system/oscar-website/pull/462, but that forgot event links at some other places.


HereAround (approved):
Looks good to me. Thank you!

PR #475, opened 2025-03-26
Improve docs on building website locally

lgoettgens:
Taken from https://github.com/gap-system/GapWWW/blob/ef6122f1f0526bb1cb43fd32a9a651d0c7e12562/README.md?plain=1#L24C1-L41C64 as suggested by @fingolfin


HereAround (approved):
Looks good to me. Thank you!

PR #480, opened 2025-03-31
Add bigdecimal to Gemfile

fingolfin:
It seems to be needed over here


aaruni96 (approved):
Just got the warning in CLI to add this. Adding this makes the warning go away, and doesn't seem to break anything else.

PR #485, opened 2025-04-11
Hot fix for error in update script

HereAround:
This should should hopefully fix the current failure in the update script.

cc @fingolfin @aaruni96


fingolfin (commented):
Looks good to me. I don't know if @aaruni96 has time to look at this?

aaruni96 (approved):
Haven't tried it out, but looks reasonable

HereAround:
Due to the time shift, I doubt if @aaruni96 could look at it now.

I will merge, so we can see in a few minutes if this serves as hotfix. And once @aaruni96 finds time, we can discuss and improve.

HereAround:
(Well, clearly I am wrong. Thank you @aaruni96.)

PR #486, opened 2025-04-15
Rewrite in python / add fallback for network failures

aaruni96:
Rewrites all my clever curls into more standard python/PyGithub, with try/catch blocks for all network requests. This should fail gracefully the next time there is a network error, and prevent more symptoms prompting #485


HereAround (approved):
Thank you for working on this @aaruni96 . Looks good to me. As per usual - best to try it out. Will thus merge this now.

PR #499, opened 2025-05-13
Change website to use the new OSCAR logo

fingolfin:
The OSCAR PIs have recently decided to replace the OSCAR logo with a revised version. This PR implements this for the website.

The website uses the SVG version of the logo, but I also provided a vector PDF & a PNG here for people who wish to e.g. use this on slides.

CC @fieker @micjoswig @simonbrandhorst @thofma 
CC @gfourier (I don't know the GitHub handle of Moritz Weber, if he has one)


fingolfin:
Actually, I forgot to update the favicons. Will do that later / tomorrow.

lgoettgens:
Is there also a version that works on a yellow background, i.e. to be used instead of https://github.com/oscar-system/book.oscar-system.org/blob/gh-pages/public/OSCAR_logo.png on book.oscar-system.org?

joschmitt:
> Is there also a version that works on a yellow background, i.e. to be used instead of https://github.com/oscar-system/book.oscar-system.org/blob/gh-pages/public/OSCAR_logo.png on book.oscar-system.org?

I can mess around with GIMP again if you want :)

fingolfin:
It should be easy enough to edit via Inkscape or even directly in the SVG.

We can also just put it on a white rounded rectangle background

ulthiel:
Ah nice, A5 Dynkin diagram.

PR #503, opened 2025-05-20
Ak96/rewrite in python attempt 2

aaruni96:
A better attempt of rewriting the update stuff from bash into python.

This requires PyGithub to be installed on the server.

The API_KEY is defined in the systemd unit file for the website update : `~/.config/systemd/user/oscar-website.service`

When a network failure occurs, the behavior is different based on what we are updating:

- Tutorial status: keeps file unchanged
- Tutorial last updated date: uses fallback date (01-01-1970) for the tutorials it couldn't fetch date for, new date for those it could fetch
- OSCAR release status: keeps file unchanged


etc/update.sh
```diff
@@ -15,24 +15,17 @@ bundle config set --local path 'vendor/bundle'
 bundle install
 
 # get tutorial status
-statusurl=$(curl -SsL --retry 5 "https://api.github.com/repos/oscar-system/TutorialTesterforOscar/actions/workflows/CI.yml/runs?per_page=1" | jq  '.workflow_runs[0].jobs_url' | sed 's/"//g')
-curl -SsL --retry 5 $statusurl | jq '.["jobs"][1:] | .[] | .name+":"+.conclusion' | sed 's/^.*\s.*\s.*\s//' | sed 's/):/: /' | sed s'/"$//' > _data/examples_status.yml
+echo "Getting tutorial status....."
+./etc/tutorial_status.py
```
fingolfin: You are now executing a Python script that has prerequisites. So this `update.sh` script should ensure those are actually satisfied / present. I.e. we need the analogue to `bundle install` above.

For GAP we do this in our CI when we need to run a Python script:
```
python -m pip install -r dev/releases/requirements.txt
```
where `dev/releases/requirements.txt` is plain text file with each line a package name.

Perhaps it should also pass `--user` to `pip` and / or should be setting up / use a Python venv .

etc/update.sh
```diff
@@ -15,24 +15,17 @@ bundle config set --local path 'vendor/bundle'
...
-echo "day: \"$day\"" >> _data/release.yml
-echo "date: \"$year-$month-$day\"" >> _data/release.yml
-rm latest.json
+echo "Getting OSCAR version info......."
+./etc/update-latest-release.py
+echo "Done!"
 # run jekyll
-bundle exec jekyll build --config _config.yml,_config_production.yml -d /srv/www/www-mathe-oscar/data/http
+echo "Running jekyl......."
+bundle exec jekyll build --config _config.yml,_config_production.yml # -d /srv/www/www-mathe-oscar/data/http
```
fingolfin: why this change? is it intentional?

etc/update-latest-release.py
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env python3
+
+import os
```
fingolfin, line 3:
```suggestion
import os
import sys
```

etc/update-latest-release.py
```diff
@@ -0,0 +1,41 @@
...
+
+from datetime import datetime
+from github import Github, Auth
+
+API_KEY = os.getenv("API_KEY")
+if API_KEY == None:
+    print("API key was not found! Authentication will fail!\nSet the environment variable API_KEY "
+          "to a github access token and try again!\n")
+
+assert API_KEY != None
```
fingolfin: It seems weird to catch an error, print a nice error message, but then proceed and instead rely on an `assert` to throw an error?
```suggestion
    print("API key was not found! Authentication will fail!\nSet the environment variable API_KEY "
          "to a github access token and try again!\n")
    sys.exit(1)
```

etc/update-latest-release.py
```diff
@@ -0,0 +1,41 @@
...
+failed=False
+
+try:
+    repo = g.get_repo("oscar-system/Oscar.jl")
+    release = repo.get_latest_release()
+except Exception as e:
+    print(e)
+    print("Network access failed!")
+    print("Leaving the release file unchanged!")
+    os.exit()
```
fingolfin: I don't think there is an `os.exit`. Did you perhaps mean this?
```suggestion
    sys.exit()
```

etc/update-latest-release.py
```diff
@@ -0,0 +1,41 @@
...
+if not failed:
+    dt = release.created_at
+    version = release.title[1:]
+
+releasestring = f"version: '{version}'\nyear: '{dt.year}'\nmonth: '{dt.month}'\nday: '{dt.day}'\ndate: {dt.date()}\n"
+
+if os.getcwd().split('/')[-1] == 'etc':
+    releasefilepath = '../_data/release.yml'
+else:
+    releasefilepath = "_data/release.yml"
```
fingolfin: This is very brittle. I would suggest we either require this path as an argument to the script; or else let the script locate itself (`os.path.abspath(sys.argv[0])`) and then take a path relative to that.

etc/update.sh
aaruni96: No, something carried forward from the last PR and I should've caught.

etc/update-latest-release.py
aaruni96: `sys.exit(1)` (and also a failing assert) will stop the update.sh script.

Should we use the builtin `exit()`, so that we exit from the python script, but the update script keeps going ?

etc/update-latest-release.py
aaruni96: I meant to use the builtin exit function, but my brain managed to convince me in the moment that I should just replace `sys` to `os`

etc/update.sh
aaruni96: I have added a venv activation phase to the script, but this will require us to have a one time creation of the venv manually before this PR is merged.

etc/update-latest-release.py
aaruni96: Does the commit https://github.com/oscar-system/oscar-website/pull/503/commits/9ce5ba3cd202f5e248a4c08d7f1bee96637c67f7 properly implement what you suggested ?

etc/update.sh
```diff
@@ -14,25 +14,27 @@ cat /srv/www/www-mathe-oscar/data/webhook.secret >> .htaccess
 bundle config set --local path 'vendor/bundle'
 bundle install
 
+# use the venv defined in .venv
+# set this up before running this script!
+echo "Activating venv..."
```
fingolfin: can't you add a line before this doing roughly `if .venv is not an existing dir, set up a venv` ?

That will make it that much easier to set up a new web server if we ever have to. The alternative would be to add this as another step in `etc/README.md` (I think adding it to the script is preferable to that)

etc/update-latest-release.py
fingolfin: Fine by me

etc/update-latest-release.py
fingolfin: it looks plausible. but of course it should be tested :-)

etc/tutorial_status.py
```diff
@@ -0,0 +1,50 @@
...
+    print("Trying to get jobs.......")
+    # anything that calls out to a network, and might face a transient error
+    repo = g.get_repo("oscar-system/TutorialTesterforOscar")
+    workflow = repo.get_workflow("CI.yml")
+    run = workflow.get_runs()[0]    # get most recent run
+    jobs = run.jobs()
+    print("Done!")
+except Exception as e:
+    print(e)
+    falied = True
```
fingolfin:
```suggestion
    failed = True
```

etc/tutorial_status.py
```diff
@@ -0,0 +1,50 @@
...
+        status = i.conclusion
+        resultstring += f"{name}: {status}\n"
+else:
+    print("Network access failed, resulting file will be unchanged!")
+
+datapath = '/'.join(os.path.abspath(sys.argv[0]).split('/')[0:-2])+'/_data'
+statusfilepath = f"{datapath}/examples_status.yml"
+
+if resultstring=="":
+    # if resultstring is empty, just use the old data
```
fingolfin: but we are not using the old state, are we? we are overwriting it now with the content of `example_status.yml` which may be old ?

Moreover, if `resultstring` is *not* empty, we are now discarding it, that seems wrong?

So maybe
```suggestion
if len(resultstring) > 0:
```


But once you do all this, couldn't you remove the `try/except`, the variable `failed` etc. -- and then the script just aborts (doing nothing) if there is an error.

In in `update.sh`, just ignore the exit code

etc/update.sh
fingolfin: As suggested elsewhere, I don't see why you can't also init the venv here if it is not yet there? Or even just init it unconditionally?

etc/update.sh
```diff
@@ -14,25 +14,27 @@ cat /srv/www/www-mathe-oscar/data/webhook.secret >> .htaccess
...
+
+# install requirements for the python scripts
+echo "Installing python pre requisites..."
+python3 -m pip install -r etc/requirements.txt
+
 # get tutorial status
-statusurl=$(curl -SsL --retry 5 "https://api.github.com/repos/oscar-system/TutorialTesterforOscar/actions/workflows/CI.yml/runs?per_page=1" | jq  '.workflow_runs[0].jobs_url' | sed 's/"//g')
-curl -SsL --retry 5 $statusurl | jq '.["jobs"][1:] | .[] | .name+":"+.conclusion' | sed 's/^.*\s.*\s.*\s//' | sed 's/):/: /' | sed s'/"$//' > _data/examples_status.yml
+echo "Getting tutorial status....."
+./etc/tutorial_status.py
```
fingolfin: To ignore the exit code, could try this:
```suggestion
./etc/tutorial_status.py || :
```

etc/update-latest-release.py
```diff
@@ -0,0 +1,41 @@
...
+try:
+    repo = g.get_repo("oscar-system/Oscar.jl")
+    release = repo.get_latest_release()
+except Exception as e:
+    print(e)
+    print("Network access failed!")
+    print("Leaving the release file unchanged!")
+    exit()
+
+if not failed:
```
fingolfin: I think `failed` is redundant now in this script and could be removed.

etc/tutorial_status.py
aaruni96: Sorry, this was supposed to be `resultstring!=""` (and your suggestion of checking length is even better).

I think the idea of `try/except` was to get a good error message. But thinking about it.., we can keep the `try/except`, and move the pretty error message to the `except` part, and then exit. This would let us get rid of the `failed` variable.

aaruni96: Does commit 7713298d look alright ?

fingolfin: It looks plausible.

but I don't underestimate why you then still need to check if resultstring is non-empty?

etc/tutorial_status.py
```diff
@@ -0,0 +1,47 @@
...
+
+resultstring = ""
+
+for i in jobs:
+    if i.name == "Prepare Tests":
+        continue
+    print(f"Processing {i.name}.....")
+    name = i.name.split()[-1][0:-1]
+    status = i.conclusion
+    resultstring += f"{name}: {status}\n"
```
fingolfin: For extra resilience I would suggest to generate YAML using quoted key/value pairs.
```suggestion
    resultstring += f"'{name}': '{status}'\n"
```

etc/update.sh
```diff
@@ -14,25 +14,31 @@ cat /srv/www/www-mathe-oscar/data/webhook.secret >> .htaccess
...
-echo "year: \"$year\"" >> _data/release.yml
-echo "month: \"$month\"" >> _data/release.yml
-echo "day: \"$day\"" >> _data/release.yml
-echo "date: \"$year-$month-$day\"" >> _data/release.yml
-rm latest.json
+echo "Getting OSCAR version info......."
+./etc/update-latest-release.py
+echo "Done!"
 # run jekyll
+echo "Running jekyl......."
```
fingolfin:
```suggestion
echo "Running jekyll......."
```
Instead of all these manuals print statements it may be more effective to add `set -x` at the start of the script

etc/tutorial_status.py
aaruni96: To guard against my paranoia for something that has very little chance of happening :

What we do in this script, is fetch the job details of the latest CI run of the tutorial tester repository ( https://github.com/oscar-system/TutorialTesterforOscar/ ). The CI run of tutorial tester fetches a job list dynamically, and it is possible that something goes wrong there, and it returns an empty list. In this case, the CI job would appear to run successfully with 0 entries. So, when we fetch this particular job, the `try/except` will succeed, and resultstring will still be empty.

I know, very unlikely to happen, but the possibility of this happening popped into my head while I was writing the script.

etc/tutorial_status.py
```diff
@@ -0,0 +1,47 @@
...
+
+for i in jobs:
+    if i.name == "Prepare Tests":
+        continue
+    print(f"Processing {i.name}.....")
+    name = i.name.split()[-1][0:-1]
+    status = i.conclusion
+    resultstring += f"'{name}': '{status}'\n"
+
+datapath = '/'.join(os.path.abspath(sys.argv[0]).split('/')[0:-2])+'/_data'
```
fingolfin: Cleaner: 
```suggestion
ownpath = os.path.abspath(sys.argv[0])
repopath = os.path.dirname(os.path.dirname(ownpath))
datapath = repopath +'/_data'
```
... possibly for a future update

etc/update-dates.py
```diff
@@ -1,37 +1,49 @@
...
+
+g = Github(auth=auth)
+failed=False
 
-    # fallback default date
-    dt = datetime(1970, 1, 1)
+if os.getcwd().split('/')[-1] == 'etc':
+    ogfile = yaml.safe_load(open("../_data/examples.yml"))
+else:
+    ogfile = yaml.safe_load(open("_data/examples.yml"))
```
fingolfin: like in the other file this hack should be replaced

fingolfin:
Since website updates are basically completely broken right now, we need to move quickly with this.

I'd be OK with merging this now and fixing it later, but as long as we set no `API_KEY` it won't do much. But perhaps that good...

But if this fails to work again, I'd rather rip it all out for now (well, deactivate it in the `update.sh` script).


Overall I'd feel much better if I knew these scripts were tested at all; either by you, locally, or (much better on the long run) in a GitHub Action -- so that e.g. syntax errors or typos in variables names are caught before we try it in the wild. But this can come later...

aaruni96:
> but as long as we set no API_KEY it won't do much

I set the API_KEY in the systemd unit file (mentioned at the top of this PR). I will also add this to the readme.

PR #519, opened 2025-06-11
Add another redirect

HereAround:


aaruni96:
Should fix the problem discovered in Triage today, about the link in Oscar's `LICENSE.md` pointing to the right page on the website.

Thanks @HereAround

fingolfin:
It works, thank you! Maybe do the same for `credits` a line above?

aaruni96:
I don't understand why yet, but credits already redirects to the desired page, both with and without the slash (alteast on my testing with firefox and chromium and curl...).

PR #521, opened 2025-06-12
Additional redirect for credits and related questions

HereAround:
As requested by @fingolfin in https://github.com/oscar-system/oscar-website/pull/519, we could add another redirect for credits (without backslash). This is the one-line change in this PR.

This touches on https://github.com/oscar-system/oscar-website/issues/517. To understand this issue, please go to `https://www.oscar-system.org/` and `click` on credits in the sidebar. I then see the following:

Picture2

I.e. the click on credits immediately takes me to the software credits. This is not in tandem with the side layout and setup for the following reasons:
1. If clicking on credits takes us to a submenu immediately, then probably it should be the very first submenu and not the second, right? This can for instance be fixed by changing the order of the submenus. (On second thought, adjusting the credit redirect to take us to the people credits is not a good idea. Likely, credits is used in many other places - say the OSCAR documentation - to give credits to the underlying software that OSCAR build upon.)

2. We have a dedicated credits page, which is not accessible due to our redirects (or so I believe). When running jekyll on my computer, I can access it. Here is what this page looks like:

Picture

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -19,5 +19,6 @@ RedirectMatch temp ^/getting-started/tutorials/(.*) https://www.oscar-system.org
 RedirectMatch temp ^/getting-started/documentation/(.*) https://www.oscar-system.org/documentation/$1
 RedirectMatch temp ^/community/(.*) https://www.oscar-system.org/contact-and-support/$1
 RedirectMatch permanent ^/credits/$ https://www.oscar-system.org/credits/software-credits/
+RedirectMatch permanent ^/credits$ https://www.oscar-system.org/credits/software-credits/
 RedirectMatch permanent ^/contributors/$ https://www.oscar-system.org/credits/contributors/
 RedirectMatch permanent ^/contributors$ https://www.oscar-system.org/credits/contributors/
--- a/_data/sidebar.yml
+++ b/_data/sidebar.yml
@@ -40,10 +40,10 @@
 - title: "Credits"
   url: "/credits/"
   submenu:
+    - title: "Software"
+      url: "/credits/software-credits/"  
     - title: "People"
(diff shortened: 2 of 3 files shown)


HereAround:
I have just pushed one more commit: https://github.com/oscar-system/oscar-website/pull/521/commits/108036ac43f1fe758e9c50e71619237d545921f8 which shuffles the submenus within credits.

Regarding the index page within credits, I am not sure how we want to deal with it. Just delete it? Have it link to the software credits? Something else?

@aaruni96 @fingolfin what do you think?

aaruni96:
If I remember correctly, we added the redirect from credits to software-credits to preserve backwards compatibility. So, for as long as we want to preserve that, it cannot redirect to any place but the software credits. In any case, we should definitely make it redirect to the first entry of the submenu, so, I am in favor if moving "Software" above "People" if we want to preserve the current redirect.

HereAround:
> If I remember correctly, we added the redirect from credits to software-credits to preserve backwards compatibility. So, for as long as we want to preserve that, it cannot redirect to any place but the software credits. In any case, we should definitely make it redirect to the first entry of the submenu, so, I am in favor if moving "Software" above "People" if we want to preserve the current redirect.

I was wondering if we want to keep the index page in the credits folder. As I see it, this index page is not accessible, due to the redirects.

aaruni96:
I would say, lets keep it. It doesn't hurt (since the redirects make sure people never see it), but if for whatever reason in the future, the website is deployed somewhere else (or, for example, testing on localhost), clicking on "Credits" on the sidebar doesn't lead to a 404 page.

HereAround:
Then this PR is ready for merge.

PR #524, opened 2025-06-13
Changes to make `Installation` on the sidebar be bold 

### LINKED ISSUE #522: The word "Installation" in the sidebar is not displayed in bold font upon clicking it
Image

aaruni96:
Changes to make `Installation` on the sidebar be bold  when viewing an installation page.

This involves changing many URLS from `install-platform` to `install/platform` , and might require updating some links in other places in the Oscar universe like documentation, etc. (Although, we really shouldn't be linking to platform specific install pages from outside the website itself, no ?)

This change also hard codes `home` to be in the sidebar.html, and not generated from sidebar.yml. This is because the logic which makes the active sidebar item appear bold checks for "if item URL occurs in current page URL". Since home is `/`, and `/` appears in every URL, this would make home be bold all the time. Instead, for just that one we check "if item URL is baseURL".

Fixes #522

### DIFF
--- a/_data/sidebar.yml
+++ b/_data/sidebar.yml
@@ -1,6 +1,3 @@
-- title: "Home"
-  url: "/"
-
 # SPACER
 - spacer: true
 
--- a/_includes/sidebar.html
+++ b/_includes/sidebar.html
@@ -9,12 +9,16 @@
     </div>
 
     <nav class="sidebar-nav">
-
+      <div class="sidebar-nav-item {% if page.url == '/' %}active{% endif %}">
+        <a href="{{site.baseurl}}/" class="sidebar-nav-link">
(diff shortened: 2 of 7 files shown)


HereAround (approved):
I discussed these changes with @aaruni96 before the triage meeting (i.e. about one hour ago). As per usual, there are several ways to achieve what one desires. While this solution works, it required changing a number of URLs. Specifically, the URLs to OS-specific installation instructions. We expect that links from other pages (say the OSCAR documentation) do not link to these OS specific pages but the generic installation page, which then redirects according to the OS. So this should not be a problem. Or hopefully so.

With that being said - looks good to me and works for me (locally). Therefore, I will merge this now.

PR #526, opened 2025-06-16
Hotfix: add trailing slash to fix windows install page URL

aaruni96:
Fixes a problem raised in slack, where `https://oscar-system.org/install/win/` doesn't work, but `https://oscar-system.org/install/win` does work. (Now, both should work)

I am merging this already because currently people get a 404 when trying to look at windows install instructions, but @HereAround feel free to revert this if you think this is wrong.

### DIFF
--- a/install-win.md
+++ b/install-win.md
@@ -1,6 +1,6 @@
 ---
 layout: page
-permalink: /install/win
+permalink: /install/win/
 ---
 
 <div class="platform-tabs">


HereAround:
Thank you @aaruni96 - I believe I verified this on your PR last week and was until now under the impression that it worked locally for me. Thus, am surprised to see this failure now. In any case - thank you for the (hot) fix.

PR #534, opened 2025-06-23
Trim requirements file

aaruni96:
Trim requirements file to not pin automatic / recursive imports to a specific version.

All libraries we explicitly import are now specified in the `requirements.txt` file, and each package then specifies its own version constraints.

This will decrease the number of PRs like #530 , and these libraries will be auto updated (within the constraints we require) whenever the script is run.

### DIFF
--- a/etc/requirements.txt
+++ b/etc/requirements.txt
@@ -1,15 +1,3 @@
-certifi==2025.4.26
-cffi==1.17.1
-charset-normalizer==3.4.2
-cryptography==45.0.3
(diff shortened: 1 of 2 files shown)


HereAround:
Thank you @aaruni96 . Looks good to me.

HereAround (approved):

HereAround:
@fingolfin Any thoughts? If not, I shall merge this PR in the afternoon tomorrow (Friday, June 27). Thank you!

PR #541, opened 2025-07-09
Add aka to data, update script accordingly

aaruni96:
Also, rename "guy" to "people", be slightly more inclusive


aaruni96:
Standing over my shoulder, @HereAround verbally approved these changes

PR #547, opened 2025-07-19
Show OSCAR logo above content when sidebar is hidden

### LINKED ISSUE #540: OSCAR logo is not visible when visiting the OSCAR website on my phone
... well at least until I open the navbar.

I think that at least on the front page we should show the OSCAR logo also on a phone at the top.

This could be done by adding an `img` to the front page at the top which is only shown when the navigation is hidden, via a suitable CSS media query.

fingolfin:
Preview at https://fingolfin.github.io/oscar-website/

Resolves #540

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -14,7 +14,7 @@
   </title>
 
   <!-- CSS -->
-  <link rel="stylesheet" href="{{ site.baseurl }}/public/css/oscar.css?version=2">
+  <link rel="stylesheet" href="{{ site.baseurl }}/public/css/oscar.css?version=3">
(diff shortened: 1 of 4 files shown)


HereAround:
I'm generally in favor of this, but I have a few comments. Here's how it looks on my mobile:

signal-2025-07-21-175125

HereAround:
- **Question 1**: Is it intentional that the red ribbon overlaps the OSCAR logo? Visually, it feels a bit off to me. My instinct is to nudge the logo left so the ribbon doesn’t sit on top of it.
- **Question 2**: It feels like there’s a bit too much "OSCAR" packed into the top of the page. The logo says OSCAR, followed by the title *"OSCAR Computer Algebra System"*, and then it appears again in the very first sentence. Maybe on mobile we could simplify things by using just the logo in place of the page title?

fingolfin:
Re 1 I suggest we just drop the ribbon completely.

Re 2 we could hide the title on the fron page in the "narrow" case.

I've updated https://fingolfin.github.io/oscar-website/ <s>(but not yet this branch)</s> to illustrate these ideas)

HereAround (approved):
Thank you @fingolfin . Looks good to me.

PR #548, opened 2025-07-23
Set white background for burger menu

fingolfin:
Before:
Screenshot 2025-07-24 at 00 18 44

After:
Screenshot 2025-07-24 at 00 17 41


HereAround (approved):
Thank you @fingolfin . Looks good to me.

PR #563, opened 2025-09-12
Update release.yml for 1.5.0

lgoettgens:
It seems that there still is no CI job doing this.


HereAround:
The website should actually fetch this change automatically unless I am mistaken. @aaruni96 ?

HereAround:
I just checked https://www.oscar-system.org/, which displays version 1.5 from September 11 for me.

Unless I am mistaken, this information is updated once we test the tutorials (daily at about 4am). This test always verifies if the tutorials work with the latest OSCAR version, thus this action must automatically figure out the latest release version.

In any case, the information on the website seems up-to-date now. Will thus close this PR.

aaruni96:
> The website should actually fetch this change automatically

This is correct. The information in the yml file is just a fallback, in case the automatic infrastructure fails catastrophically (and so that there is *some* version in there when previewing the website locally)

PR #579, opened 2025-09-16
Added Username (Tobias) to list of participants (summer school 2025)

Tobias271828:


lgoettgens (changes requested):
This is not valid yaml (aka not in the format of the other entries)

PR #633, opened 2025-10-15
Add repo to people_list

aaruni96:


etc/update-contributors.py
```diff
@@ -133,7 +139,17 @@ def write_line_break(self, data=None):
             newpersonlist.append(i[0])
             print(f"{i[0]}\t{i[1]}\t{github_username}")
             github_newusers.append(github_username)
-            newList.append([i[0], i[1], github_username])
+            newList.append([i[0], i[1], github_username, [repo]])
+        elif github_username in names:
+            print(github_username)
```
HereAround: Why is this printing necessary?

HereAround:
```suggestion
        elif github_username in names:
```

aaruni96: Debug output which sneaked past the cleaning phase.

HereAround (approved):
Thank you @aaruni96 . LGTM!

PR #648, opened 2025-10-30
Set up dependabot

fingolfin:
This helps ensure our GH Workflows don't use outdated actions


fingolfin:
I noticed that the update contributors workflow still uses "actions/checkout@v3" which is outdated. I could update this manually, but using dependabot will do it for us, and won't forget to do it... I am intentionally not updating the outdated action here so that we get a test for the dependabot setup.

PR #654, opened 2025-10-31
Fix contributor script

HereAround:
After some "surgery" I found the actual cause to be this:

`fatal: using multiple --group options with stdin is not supported`

Our actual call to receive the short log failed silently for this reason. The changes in this PR should fix this short coming. Once merged, I would eyeball https://github.com/oscar-system/oscar-website/pull/642 to gauge if the extracted information is fully correct. However, superficially, it seems alright. Cf. the log:

https://github.com/oscar-system/oscar-website/actions/runs/18977543665/job/54201147172?pr=654

### DIFF
--- a/.github/workflows/test-contributer-script.yml
+++ b/.github/workflows/test-contributer-script.yml
@@ -25,6 +25,11 @@ jobs:
         python -m pip install --upgrade pip
         pip install PyGithub pyyaml requests
 
+    - name: Prepare git
+      run: |
+        git --version
+        git config --global --add safe.directory '*'
+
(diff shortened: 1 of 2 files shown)


HereAround:
Am curious to check with https://github.com/oscar-system/oscar-website/pull/642 and to move on in case there are further failures. So merged as is now.

HereAround:
Definitely progress, but not perfect (cf. https://github.com/oscar-system/oscar-website/pull/642):
* "Normal" authors are listed as coauthors.
* Coauthors appear repeatedly.

Needs fixing in a separate PR.

PR #661, opened 2025-10-31
Refactor contributor script to next level

HereAround:
The next level contributor script:
* Logic mostly as before.
* Key change 1: Treat coauthors the same way as authors. In particular, no separate record/report in summary.
* Key change 2: We recently decided that the people_list.yml need not be sooo human readable. As such, no point in having empty lines separate records. Have thus also dropped the old dumper.

Other than that, a massive refactor. Hopefully the script is now safer to execute, catches the coauthors correctly, easier to maintain and to read.

The result of its execution can be seen here: https://github.com/oscar-system/oscar-website/actions/runs/19005559056/job/54278479324?pr=661

Specifically, it finds:

> This PR updates the contributors list based on the latest changes.
  New contributors : 4 | ['Anna Hofer', 'Fabian Lenzen', 'StardustMath', 'Erroxe']
  Newly retired contributors : 0 | []
  Summary Notes:
> - Skipping non-address line in oscar-system/Oscar.jl: 'Benjamin Lorenz'

All of those four are coauthors. Once you run the script locally (or after merging and letting the bot run it), you will see that the script found one valid commit hash for each.

(I just notice that I dropped the "revived authors" comment, i.e. it no longer says "author x was revived". If I recall correctly there is exactly one such person as of recent. This can of course and should be added back in...)

cc @aaruni96

### DIFF
--- a/etc/update-contributors.py
+++ b/etc/update-contributors.py
@@ -67,7 +67,28 @@ def custom_sort_function(item):
 
 
 ##########################################
-# 2. Read information from people_list.yml
+# 2. Global variables
+##########################################
+
+# Step 3: The information on contributors, that is currently saved in PEOPLE_LIST_FILE.
+#current_contributors
+
+# Step 4: People come with alias emails. This dict will tell us the owner of a given email, provided our PEOPLE_LIST_FILE knows about this
+email_owner = {}  # lowercased email -> person dict
+
+# Step 4: Same as email_owner, but with name alias instead
+name_owner  = {}  # normalized name -> person dict
+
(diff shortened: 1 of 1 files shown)


HereAround:
I just added the revived logic back in as well, see the summary.txt output in https://github.com/oscar-system/oscar-website/actions/runs/19005671944/job/54278746142?pr=661.

As I see it, this is now ready to be merged @aaruni96 .

HereAround:
Am rather curious to see this in action. So will merge this now to investigate further @aaruni96

PR #688, opened 2025-11-10
Spell out month names fully (so August instead of Aug)

fingolfin:
Using names like Aug or Jul for months is maybe OK for
long time programmers, but I think for most "normal"
people it looks strange.


aaruni96 (approved):
I am indifferent to either way, and therefore, submit to your wisdom.

PR #691, opened 2025-11-10
etc/update-latest-release.py: some tweaks

fingolfin:
- include the julia-min version
- adjust output to be closer to what is in the repo
- use a multiline string for enhance readability


HereAround (approved):
LGTM. Thank you!

PR #696, opened 2025-11-12
Fix bug in events

HereAround:


_data/events.yml
```diff
@@ -1,4 +1,4 @@
-- title: albatross-2026: ALgeBrAic meThods foR pOlynomial System Solving 2026
+- title: albatross-2026
```
fingolfin, line 1: You could also put quotation marks around the text, then a colon works fine

PR #710, opened 2025-11-19
Display a few conferences only

### LINKED ISSUE #706: Too many upcoming events on the front page also isn't good...
https://www.oscar-system.org now lists a lot lot *lot* of news. More than an entire page for me.

That's IMHO too much... Thanks for collecting them, but we should not show them all *at once*.

I suggest we only show the next N upcoming events (for N something like 3 or 4 maybe) and then a "More..." link (*if* there are more) that goes to, say, <https://www.oscar-system.org/events/> -- and then add the full list there?

HereAround:
Fixes https://github.com/oscar-system/oscar-website/issues/706.

### DIFF
--- a/_data/config.yml
+++ b/_data/config.yml
@@ -1,0 +1,1 @@
+number_of_displayed_events_on_index_page: 4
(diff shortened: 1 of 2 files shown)


fingolfin:
Thanks Martin!

Issue #713, opened 2025-11-19
Add reference to zenodo


HereAround:
Cf. https://github.com/oscar-system/Oscar.jl/issues/5219.

Issue #690, opened 2025-11-10
Enhance CI to update `_data/release.yml` directly or via PR


fingolfin:
So we have this workflow `.github/workflows/latest_replease.yml` [*sic*, typo fix in PR #689] which runs the `etc/update-latest-release.py` script whenever we update the `gh-pages` branch.

And then we run the `etc/update-latest-release.py` script *on the webserver* as part of the `etc/update.sh` whenever a change is pushed.

Only this last part actually does something permanent, the GH workflow however just runs a script and discards the output.

I'd like us to do this differently: `update.sh` should not call `update-latest-release.py` at all. Instead the GH workflow should, after running it, check if there were changes, and then either open a PR or push the changes to `gh-pages`. (Both have advantages; I kinda like the PR route so that a human is in the loop, but we could also directly push, that would be a bit closer to what happens now).

This way, the repository data more closely matches what is actually on the server.

Oh, and of course for this to be useful, the workflow should *not* be run upon pushes to `gh-pages`, but instead it should be run by a cron job, say once per day; or when there is a manual `workflow_dispatch` trigger.

Kinda how <https://github.com/gap-system/GapWWW/blob/master/.github/workflows/sync.yml> does it.

fingolfin:
Right now the website on e.g. https://www.oscar-system.org/install/mac/ still lists 1.5.0 as the latest.

Why? Because there were no updates to the website since the release, and so the script which is run on the server whenever we update the page, and which puts in the latest OSCAR version, did not run.

If instead we did it like here, it'd be there.

fingolfin:
Of course we could get some of this effect also by running a "cron job" (well, timed systemd unit) on the server to run the update scripts at least once every 24 (or 6 or whatever) hours)

aaruni96:
Something has broken, because the website indeed used to rebuild itself every 24 hours, but for whatever reason, has not done so since 19 November (last updated date on the website) ?

aaruni96:
> Only this last part actually does something permanent, the GH workflow however just runs a script and discards the output.

This was @HereAround's attempt to CI test the script, in the direction to address #514 . From talking to him, it was supposed to be a step1: just run the script whenever the website contents are updated, and if it doesn't run successfully, the CI action will show red.

HereAround:
@aaruni and I just discussed this on slack. Long story short: Yes, we will improve along your suggestions @fingolfin , but take it in places one step further.

Specifically:
[...]

PR #715, opened 2025-11-25
Overhaul release update and checklist mechanics

### LINKED ISSUE #612: Send email to OSCAR developer mailing list whenever we have a new OSCAR release
@fingolfin Just mentioned this to me - while the website informs on the latest OSCAR version, we should also announce a new release by sending a message to the OSCAR developer mailing list (and maybe also the newsletter should announce this?).

The goal should be to automatize this step as much as possible. My immediate thought:
* Set up a CI action here that checks if the OSCAR version changes.
* If it does change, open an issue in the website repository.
* This issue could include an auto generated text for the announcement.
* The issue should cite the relevant people and remind them to send out a corresponding email.
* The relevant people could include the following individuals: @fingolfin , @aaruni96 , @lgoettgens, myself (@HereAround).

HereAround:
Following: https://github.com/oscar-system/oscar-website/issues/690

Might close https://github.com/oscar-system/oscar-website/issues/612.

We could also take a first stab at https://github.com/oscar-system/oscar-website/issues/713, by adding a corresponding reminder to the checklist on this behalf.


HereAround:
Action triggered on demand: https://github.com/oscar-system/oscar-website/actions/runs/19673471092/job/56348132333

Opened PR: https://github.com/oscar-system/oscar-website/pull/717
Opened issue: https://github.com/oscar-system/oscar-website/issues/716

Appears to be working as desired. Ready for review.

aaruni96:
This appears to create a duplicate issue, if an issue is already opened ( example, #716 appears to be a duplicate of #714 ).

etc/update-latest-release.py
```diff
@@ -2,48 +2,115 @@
...
+
 API_KEY = (os.getenv("API_KEY") or os.getenv("GITHUB_TOKEN") or "").strip()
-if API_KEY == None:
-    print("API key was not found! Authentication will fail!\nSet the environment variable API_KEY "
-          "to a github access token and try again!\n")
-    exit()
+if not API_KEY:
+    print(
+        "API key was not found! Set API_KEY or GITHUB_TOKEN to a GitHub token.",
+        file=sys.stderr,
```
aaruni96: Do we need to specify writing the output to stderr ? That is, do we gain anything by it ?

etc/update-latest-release.py
```diff
@@ -2,48 +2,115 @@
...
 
+TARGET_REPO = os.getenv("TARGET_REPO", "oscar-system/Oscar.jl")
+JULIA_MIN = os.getenv("JULIA_MIN", "1.10")
+
+# repo root = parent of this file's directory
 ownpath = os.path.abspath(sys.argv[0])
 repopath = os.path.dirname(os.path.dirname(ownpath))
-datapath = repopath +'/_data'
+datapath = os.path.join(repopath, "_data")
+os.makedirs(datapath, exist_ok=True)
```
aaruni96:
```suggestion
```

We should not be making a directory at `datapath` here. Either it already exists (because its part of the git repo, and it has the data in it), or it doesn't (and we create it, and then we don't find the data in it, because we just created a new directory)

etc/update-latest-release.py
```diff
@@ -2,48 +2,115 @@
...
     release = repo.get_latest_release()
 except Exception as e:
-    print(e)
-    print("Network access failed!")
-    print("Leaving the release file unchanged!")
-    exit()
+    print(e, file=sys.stderr)
+    print("Network access failed! Leaving the release file unchanged!")
+    gh_output(changed="false", reason="fetch-failed")
+    sys.exit(0)
```
aaruni96: Usually, `sys.exit(0)` (exit with the return code 0), is used to denote that everything went fine with the script, and non zero return codes are error codes.

If there isn't a reason (maybe some other action depends on this action "succeeding" even when network access is failed), then we should change this to a non zero code.

etc/update-latest-release.py
```diff
@@ -2,48 +2,115 @@
...
+
+# Determine version string (strip leading 'v')
+tag = release.tag_name or release.name or release.title
```
aaruni96: Are we sure that one of these always has the version name?

On the other hand, shouldn't it be guaranteed that tag name has the version? Why do we check tag name, and then two other fall backs ?

etc/update-latest-release.py
```diff
@@ -2,48 +2,115 @@
...
+version = (tag or "").lstrip("v").strip()
 
-dt = release.created_at
-version = release.name[1:]
+if not version:
+    print("Could not determine version from latest release.", file=sys.stderr)
+    gh_output(changed="false", reason="no-version")
+    sys.exit(0)
```
aaruni96: Same comment as above about non zero exit codes.

etc/update-latest-release.py
```diff
@@ -2,48 +2,115 @@
...
 
-releasestring = f"""\
-version: "{version}"
+dt = release.published_at or release.created_at
+if not dt:
+    print("Could not determine release date.", file=sys.stderr)
+    gh_output(changed="false", reason="no-date")
+    sys.exit(0)
```
aaruni96: Ditto, about non zero exit codes.

etc/update-latest-release.py
```diff
@@ -2,48 +2,115 @@
...
-dt = release.created_at
-version = release.name[1:]
+if not version:
+    print("Could not determine version from latest release.", file=sys.stderr)
+    gh_output(changed="false", reason="no-version")
+    sys.exit(0)
 
-releasestring = f"""\
-version: "{version}"
+dt = release.published_at or release.created_at
```
aaruni96: Do we need this fallback? Shouldn't some API guarantee which entry always has the information we want ?

etc/update-latest-release.py
```diff
@@ -2,48 +2,115 @@
...
+if not dt:
+    print("Could not determine release date.", file=sys.stderr)
+    gh_output(changed="false", reason="no-date")
+    sys.exit(0)
+
+dt = dt.replace(tzinfo=timezone.utc)
+
+# Read old version (if file exists) without adding a YAML dependency
```
aaruni96: Is there a reason we insist on not adding yaml as a dependency?

With yaml, the following block turns into

```python
with open(releasepath, "r") as yamlfile:
    data = yaml.safe_load(yamlfile)
    old version = data['version']
```

HereAround: No objection against using yaml.

etc/update-latest-release.py
HereAround: I might be overly careful here. Not sure to what extend there is a guarantee. Should investigate.

etc/update-latest-release.py
HereAround: I might be overly careful. Need to investigate in more detail to which extend there is a guarantee.

etc/update-latest-release.py
HereAround: I suppose we can also just print it in the log. For us, no immediate gain (I think).

HereAround:
@aaruni96 Thank you for the feedback. Will overhaul the python script along your suggestions.

etc/update-contributors.py
```diff
@@ -37,7 +37,7 @@
              "Nemocas/Nemo.jl", "oscar-system/Oscar.jl", "oscar-system/Polymake.jl", "oscar-system/Singular.jl")
 
 # Bots we ignore
-BOT_TOKENS = ("github-actions[bot]", "dependabot[bot]", "renovate[bot]", "changelog[bot]", "update_contributors[bot]")
+BOT_TOKENS = ("github-actions[bot]", "dependabot[bot]", "renovate[bot]", "changelog[bot]", "update_contributors[bot]", "update_release[bot]")
```
aaruni96: Maybe instead of having a concrete list for bots, we can just check if a name ends with `[bot]`, and skip it (of course, after complaining about it in the log or in the summary file).

etc/update-latest-release.py
```diff
@@ -1,49 +1,111 @@
...
+def gh_output(**kvs):
+    path = os.getenv("GITHUB_OUTPUT")
+    if not path:
+        return
+    with open(path, "a", encoding="utf-8") as f:
+        for k, v in kvs.items():
+            v = "" if v is None else str(v)
+            f.write(f"{k}={v}\n")
+
+# 4. Get/set API key or raise error
```
aaruni96: Maybe this should go above, with the constants, before we start defining functions ?

etc/update-latest-release.py
```diff
@@ -1,49 +1,111 @@
...
+JULIA_MIN = os.getenv("JULIA_MIN", "1.10")
+ownpath = os.path.abspath(sys.argv[0])
+repopath = os.path.dirname(os.path.dirname(ownpath))
+datapath = os.path.join(repopath, "_data")
+releasefilepath = os.path.join(datapath, "release.yml")
+
 
+# 3. Function to write simple key=value outputs for GitHub Actions.
+def gh_output(**kvs):
+    path = os.getenv("GITHUB_OUTPUT")
```
aaruni96, line 32: In a previous commit ( 96ffbe6778e43027f2474c787b3260df94eb43ca ), this had a docstring. Why did we remove it ?

etc/update-latest-release.py
```diff
@@ -1,49 +1,111 @@
...
+    sys.exit(1)
 
-dt = release.created_at
-version = release.name[1:]
 
-releasestring = f"""\
-version: "{version}"
+# 6. Process release version
+tag = release.tag_name
+if not tag: # Safety fallbacks (rarely needed but harmless)
```
aaruni96: We should get rid of this if clause, IMO.

An API is a promise and a guarantee of "if you ask for this endpoint, you will get this information".

If we expect the answer to be in the `tag_name` field, and we find something unexpected when we look, either (a) the OSCAR release process has failed and included garbage (or emptiness) where it shouldn't have, or (b) the API has lied to us

In either case, I would want an error to investigate, instead of trying to look in other places.

etc/update-latest-release.py
```diff
@@ -1,49 +1,111 @@
...
+    tag = release.name or release.title
+version = (tag or "").lstrip("v").strip()
```
aaruni96:
```suggestion
version = (tag or "").strip().lstrip("v")
```
As long as we are being safe, we might as well be really safe. `strip()` strips whitepsace from both ends, and then `lstrip()` can strip the v at the beginning.

(Except if you are expecting a string of the form `v<whitespace>x.y.z`, in which case, disregard my comment.)

etc/update-latest-release.py
```diff
@@ -19,67 +27,56 @@ def gh_output(**kvs):
...
-        for line in f:
-            if line.strip().startswith("version:"):
-                # version: "1.2.3"
-                _, val = line.split(":", 1)
-                old_version = val.strip().strip('"')
-                break
+    with open(releasefilepath, "r") as yamlfile:
+        data = yaml.safe_load(yamlfile)
+        old_version = data['version']
+
```
aaruni96, line 74: We may want to add an else block here and error / complain ?

aaruni96 (dismissed):

etc/update-contributors.py
aaruni96: (the suggested change should go to another PR)

.github/workflows/release-checklist.yml
```diff
@@ -12,51 +10,64 @@ concurrency:
...
+  pull-requests: write
 
 jobs:
-  release-checlist:
+  release-check:
     runs-on: ubuntu-latest
 
     steps:
       - name: Checkout repository
         uses: actions/checkout@v5
```
fingolfin:
```suggestion
        uses: actions/checkout@v6
```

.github/workflows/release-checklist.yml
```diff
@@ -12,51 +10,64 @@ concurrency:
...
         uses: actions/setup-python@v6
         with:
           python-version: "3.x"
 
-      - name: Install deps
+      - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install PyGithub
+          pip install PyGithub PyYAML
```
fingolfin:
```suggestion
          python -m pip install PyGithub PyYAML
```
though the cleaner way would be to do this via a `requirements.txt` file 
```suggestion
          python -m pip install -r etc/requirements.txt
```
with `requirements.txt` having this content:
```
PyGitHub
PyYAML
```
so that anyone who wants to use this code interactively (outside of the actions) can do use this as well.

Alas this is less important overall, :-)

etc/update-latest-release.py
```diff
@@ -1,49 +1,111 @@
...
-    repo = g.get_repo("oscar-system/Oscar.jl")
+    repo = g.get_repo(TARGET_REPO)
     release = repo.get_latest_release()
 except Exception as e:
     print(e)
-    print("Network access failed!")
-    print("Leaving the release file unchanged!")
-    exit()
+    print("Network access failed! Leaving the release file unchanged!")
+    gh_output(changed="false", reason="fetch-failed")
```
fingolfin: Why is this even needed, won't the CI job just be aborted due to the non-zero exit code in the next line?

etc/update-contributors.py
HereAround: See https://github.com/oscar-system/oscar-website/issues/723.

.github/workflows/release-checklist.yml
```diff
@@ -12,51 +10,64 @@ concurrency:
...
-        id: probe
+      - name: Update latest release file
+        id: release
         env:
           API_KEY: ${{ secrets.GITHUB_TOKEN }}
           TARGET_REPO: oscar-system/Oscar.jl
-          RECENCY_DAYS: "1"
-          ISSUE_PREFIX: "OSCAR"
-          ISSUE_LABELS: "release-process"
+          JULIA_MIN: "1.10"
```
fingolfin: TODO: this should not be hardcoded, instead, it should be computed. I.e. by inspecting the `Project.toml` of the Oscar version.

(It is OK from my POV to hardcode it now, but I would not do it here, I'd rather do it in the script together with a TODO comment)

etc/update-latest-release.py
HereAround: In principle so... but, in case we do not get a key, we want to use the function `gh_output`. That function is defined just above (until line 28). So, I would leave it where it currently is.

I do not have a strong opinion @aaruni96. If you still want to restructure, please provide a suggestion or just push to this branch.

etc/update-latest-release.py
```diff
@@ -1,49 +1,111 @@
...
+
+
+# 10. Otherwise, write new release information and signal changes to github workflow
+releasestring = f"""version: "{version}"
 year: "{dt.year}"
 month: "{dt.month}"
 day: "{dt.day}"
 date: "{dt.date()}"
-julia-min: "1.10"
+julia-min: "{JULIA_MIN}"
```
fingolfin: as mentioned above: the `JULIA_MIN` variable here gains us nothing, the real solution would be to query the minimum OSCAR version by inspecting the `Project.toml`, or even by running Julia`. This can be noted here with a TODO comment.

etc/update-latest-release.py
HereAround: I thought the comment in line 20 is enough.

etc/update-latest-release.py
```diff
@@ -1,49 +1,111 @@
...
+
+# 8. Read old version
+old_version = None
+if os.path.exists(releasefilepath):
+    with open(releasefilepath, "r") as yamlfile:
+        data = yaml.safe_load(yamlfile)
+        old_version = data['version']
+
+
+# 9. Early exist in case old version and new version agree
```
fingolfin:
```suggestion
# 9. Early exit in case old version and new version agree
```

etc/update-latest-release.py
HereAround: Ok. Done.

HereAround: Will push the change momentarily...

etc/update-latest-release.py
```diff
@@ -1,49 +1,111 @@
...
+with open(releasefilepath, "w", encoding="utf-8") as releasefile:
     releasefile.write(releasestring)
+print(f"Updated {releasefilepath} to version {version} (was {old_version or 'none'}).")
+gh_output(
+    changed="true",
+    reason="updated",
+    version=version,
+    old_version=old_version or "",
+    published_utc=dt.isoformat(timespec="seconds"),
+    target_repo=TARGET_REPO,
```
fingolfin: As far as I can tell, only `changed`, `version` and `published_utc` are used?
```suggestion
    version=version,
    published_utc=dt.isoformat(timespec="seconds"),
```

etc/update-latest-release.py
```diff
@@ -1,49 +1,111 @@
...
+# 1. Imports
 import os
 import sys
-
-from datetime import datetime
+from datetime import timezone
 from github import Github, Auth
+import yaml
 
-API_KEY = (os.getenv("API_KEY") or os.getenv("GITHUB_TOKEN") or "").strip()
```
fingolfin, line 9: Why is this getting moved downwards? Normally I'd say it is good to error out as quickly as possible

aaruni96:
I think I managed to address all the comments that were pending for this PR.

Just one thing I don't understand, why does the github diff viewer show `_data/release.yml` as a changed file ( https://github.com/oscar-system/oscar-website/pull/715/files#diff-8a15bca4f3ee985abd08ad5daa02c63cad24eb954e307de3c7827415df8d7818 ) ? The file already has the up to date information as of commit 146b402 .

fingolfin:
I think this is a bug in GitHub's PR diff viewer try merging or rebasing onto the latest gh-pages branch, that should cure it.

benlorenz:
> Just one thing I don't understand, why does the github diff viewer show `_data/release.yml` as a changed file ( [#715 (files)](https://github.com/oscar-system/oscar-website/pull/715/files#diff-8a15bca4f3ee985abd08ad5daa02c63cad24eb954e307de3c7827415df8d7818) ) ? The file already has the up to date information as of commit [146b402](https://github.com/oscar-system/oscar-website/commit/146b40201d69a5d466562476368be312027f226f) .

I think as long as the update commit is also here on the branch (`Update latest OSCAR release to v1.6.0` 8b9d623351143de6e9d60d8ca86e262233dbd494, cherry-picked? rebased?) it will show up in the diff even if the same changes have been made on master (it probably uses `merge-base` for the diff). It should disappear after something like `git rebase --empty=drop`.

aaruni96:
I think this one is ready to be merged, hoping I haven't missed anything. Could at least one more person sign off on this?

@HereAround @fingolfin @benlorenz

.github/workflows/release-checklist.yml
```diff
@@ -12,51 +10,63 @@ concurrency:
...
+        if: steps.release.outputs.changed == 'true'
+        uses: peter-evans/create-pull-request@v7
+        with:
+          commit-message: "Update latest OSCAR release to v${{ steps.release.outputs.version }}"
+          title: "Update latest OSCAR release to v${{ steps.release.outputs.version }}"
+          body: |
+            This PR updates `_data/release.yml` to the latest OSCAR release `v${{ steps.release.outputs.version }}`.
+
+            _Opened automatically by a scheduled workflow._
+          branch: "bot/update-latest-release-v${{ steps.release.outputs.version }}"
```
HereAround, line 69: We must ensure changes are made to the branch gh-pages.

Use `base: gh-pages` to do so.

HereAround, line 69:
```suggestion
          branch: "bot/update-latest-release-v${{ steps.release.outputs.version }}"
          base: gh-pages
```

HereAround:
@aaruni96 and I just discussed via slack - we are happy to merge as currently is.

However, we agreed that we definitely want to improve the CI tests: Some of those CI tests just run the python scripts. I like to view this as a first step towards a CI for the website: without doing anything I get informed if the scripts run without raising an error. Since I am often lazy and do not (wish to ) run (all?) scripts locally before merging a PR. Aaruni begs to differ. And I certainly agree that we should push this to the next level for a good website CI.

Consensus: We execute corresponding CI improvements in a future PR. Maybe @aaruni96 could you prepare a draft at some point? Not urgent from my perspective.

fingolfin:
I have some concerns about this. If a script runs without an error, then this is of course a necessary condition for it working, but by far not sufficient. I really think there should be a proper test -- esp. if you, @HereAround, indeed like to make PRs without locally testing. All kinds of things could go wrong that break the script, and we'd only notice sometime after an OSCAR, when we'd be wondering "wait, why isn't the website mentioning the new version".

Here is what I'd suggest you do: *before* running the script, delete the release.yml it is supposed to delete. Then *after* running, you can test whether the file exists; and if it does, run some code (e.g. a short python snippet) which parses the YAML and verifies that all the expected keys are present. (More could be done to verify the values are also "sane", but I don't think it's necessary; if a YAML file with suitable keys was generated we must have gotten to the very end of the script and produced something.)

And in addition, I would not do this is in a *separate* workflow; rather, I'd put this into the workflow which creates the PR (that means this workflow needs to also trigger for pull requests). I'd add an `if: CONDITION` trigger to the step which creates the PR to make sure it is *not* run when triggered by PR (or perhaps better: to make sure it is *only* run when triggered manually or by a schedule).

The advantage of that is that you also test the workflow.

PR #741, opened 2026-01-20
Improve CI

HereAround:
This aims to implement the strategy suggested in https://github.com/oscar-system/oscar-website/pull/715#issuecomment-3729935760.

In one specific instance, this PR deviates: The python script, which updates the release data, is written to fail should `_data/release.yml` be absent. Consequently, I do not delete this file before running the python script, even though suggested in https://github.com/oscar-system/oscar-website/pull/715#issuecomment-3729935760.

Update:
I am extending the scope of this PR to add similar proper tests to all four github actions that we currently have for the website.


.github/workflows/test-contributer-script.yml
```diff
@@ -40,12 +40,40 @@ jobs:
...
+          echo "ERROR: summary.txt was not created (or is empty). Failing."
+          ls -la
+          exit 1
+        fi
+
     - name: Show summary.txt
       run: |
         echo "::group::summary.txt"
-        sed -n '1,200p' summary.txt || echo "summary.txt not found"
+        sed -n '1,200p' summary.txt
```
aaruni96: Is there a reason we use `sed` here, instead of the more standard `cat`, if all we want to display the contents of `summary.txt` ?

.github/workflows/tutorial-last-modified-dates.yml
```diff
@@ -1,4 +1,4 @@
-name: Tutorial Last Modified Date Script
+name: Update Tutorial Metadata
```
aaruni96: The tutorial update script only updates the date. Is there any other metadata of the tutorials that we want to update ? Otherwise, this renaming seems confusing to me.

.github/workflows/tutorial_status.yml
```diff
@@ -1,4 +1,4 @@
-name: Tutorial Status Script
+name: Validate Tutorials
```
aaruni96: Doesn't this script just fetch the status of the tutorials, as declared by our testing repository?

In my head, to validate, implies some kind of testing is taking place..... no?

.github/workflows/test-contributer-script.yml
```diff
@@ -40,12 +40,40 @@ jobs:
         python3 update-contributors.py
         cd ..
 
+    - name: Ensure summary.txt exists
+      run: |
+        if [ ! -s summary.txt ]; then
+          echo "ERROR: summary.txt was not created (or is empty). Failing."
+          ls -la
```
aaruni96:
```suggestion
          ls summary.txt
```

Listing every file doesn't help us in a way that I can see. Maybe you want to check whether `summary.txt` was created or not? (Although, just off the top of my head, if the script itself doesn't error out, I don't see why the summary file should ever not be created, or be empty. Writing the summary file is not behind any loops or conditionals, and its content is at least partly hardcoded, so, it will never be empty.)

.github/workflows/release-checklist.yml
```diff
@@ -34,16 +35,29 @@ jobs:
           python -m pip install --upgrade pip
           python -m pip install -r etc/requirements.txt
 
-      - name: Update latest release file
+      - name: Detect new OSCAR release
```
aaruni96: I am not fully onboard with this rename. Yes, the script must first detect whether or not a new OSCAR version has been released, but then it also updates the release file in the same step. The renaming, however, makes it sound like like this script does not update the release file.

.github/workflows/release-checklist.yml
```diff
@@ -34,16 +35,29 @@ jobs:
...
         id: release
         env:
           API_KEY: ${{ secrets.GITHUB_TOKEN }}
           TARGET_REPO: oscar-system/Oscar.jl
         run: |
           python3 etc/update-latest-release.py
 
-      - name: Open release checklist issue
-        if: steps.release.outputs.changed == 'true'
+      - name: Validate _data/release.yml structure
```
aaruni96, line 49: This validation is a nice step, but more is required: 

it just checks that if we update the file, the new file still has the information we need. It does not deal with another likely situation: a new OSCAR version is released, but the script does not realise this. So, in the current workflow, such a script would be marked as "green" in the CI.

Just to be clear, I like this validation step, and this should be retained. But something more should be added to this workflow to actually check the scenario I described (and maybe more failure modes which we should foresee?).

aaruni96 (changes requested):
Thanks for the work so far @HereAround . I have left some comments. 

And, I assume, the workflows for the renamed files (`tutorial-last-modified-dates.yml`, and `tutorial_status.yml`) will be updated eventually to add some kind of checks?

.github/workflows/test-contributer-script.yml
HereAround: No reason. We can switch.

.github/workflows/tutorial-last-modified-dates.yml
HereAround: Just wanted to make the name less confusing to me and others, and this is what I came up with.

Maybe "Update Last Modified Date of all Tutorials" instead? A bit long though.

.github/workflows/tutorial_status.yml
HereAround: Yes, it indeed fetches their status and then the website sorts them accordingly.

Same as above, I found the current name confusing and was hoping to make it clearer to me and others. But then, maybe this change is not achieving this goal.

"Load status reports of all tutorials (i.e. true/false depending on whether the tutorial passed our tests)"

Again... definitely too long. But maybe something like this?

.github/workflows/test-contributer-script.yml
HereAround: I have not yet checked if the script errors itself should it ever fail to create this file.

As of right now, my thought was that it might be good to verify the file exists before we try to display its content (next step).

Regarding the `ls -la`: My thought was that if the file is not present we might want to see which files do exist. For instance, is there a .txt file but the name was somehow corrupted.

Does that make sense? If not, happy to just drop the `ls -la`.

.github/workflows/release-checklist.yml
HereAround: Fair enough.

Same principle as above - was just aiming for better names for these steps. Happy to hear better suggestions.

The reason why I am not happy with the current name "update latest release file" is twofold:
1. We do not update the file unless there is indeed a new release.
2. There is no **latest** release file, is there?

How about "Detect new OSCAR release (and update release file if needed)". A bit long, but explicit. Is that better?

.github/workflows/release-checklist.yml
HereAround: Nice to hear that you like this first step. And I agree, more can be done. Maybe should.

How could we check your scenario? The best I can right now think of is to have a check with the exact same script (or mild modifications thereof) to fetch the latest OSCAR version and then compare this again the result from the update script. But of course, this is leading to nothing, for both scripts/checks would likely both fail/succeed at the exact same time/settings to fetch the latest OSCAR version.

HereAround:
> Thanks for the work so far @HereAround . I have left some comments.
> 
> And, I assume, the workflows for the renamed files (`tutorial-last-modified-dates.yml`, and `tutorial_status.yml`) will be updated eventually to add some kind of checks?

Exactly. In the process of doing such things. Just started with checking release information, then realized the extension to the contributors is more or less similar. And so, of course, should extend to the remaining tests as well. Hope to add this shortly. (Maybe a separate PR?)

But yes, exactly. Similar tests and extensions thereof to be added hopefully soon.

.github/workflows/test-contributer-script.yml
HereAround: Done.

.github/workflows/test-contributer-script.yml
HereAround: Ok. I just removed it.

.github/workflows/tutorial-last-modified-dates.yml
HereAround: New and hopefully better name suggested in the code.

.github/workflows/tutorial_status.yml
HereAround: New and hopefully better name suggested in the code.

.github/workflows/release-checklist.yml
HereAround: New and hopefully better name suggested in the code.

.github/workflows/release-checklist.yml
HereAround: This one is still open...

What comes to my mind is that we could try to fetch the latest OSCAR release in a second independent way, i.e. without using the github API. Not sure yet how and also not if this is the best way to resolve this.

HereAround:
@aaruni96 This is ready for another round of reviewing. Hopefully I could resolve some of the issues that you found before.

* I have thought of better names again and made corresponding edits.
* Tests have been extended to all four workflows.

.github/workflows/release-checklist.yml
aaruni96: One idea is what @fingolfin originally suggested: 

> before running the script, delete the release.yml it is supposed to delete. Then after running, you can test whether the file exists; and if it does, run some code (e.g. a short python snippet) which parses the YAML and verifies that all the expected keys are present

The way I understand it is, have two separate workflows: one which tests using this strategy, and is run as part of CI, and a second one, which actually checks for a new version, i.e, does not delete the `release.yml` file first, and then makes a PR if there is a new version. The "test" workflow, can be run on every PR, and the "update" workflow can be run on a schedule. Does this sound like a good idea?

.github/workflows/tutorial_status.yml
```diff
@@ -26,10 +26,33 @@ jobs:
           python -m pip install --upgrade pip
           python -m pip install -r etc/requirements.txt
 
-      - name: Run Tutorial Status Script
+      - name: Update Tutorial Status
         env:
           API_KEY: ${{ secrets.API_KEY }}
```
aaruni96:
```suggestion
          API_KEY: ${{ secrets.GITHUB_TOKEN }}
```

.github/workflows/tutorial_status.yml
```diff
@@ -26,10 +26,33 @@ jobs:
...
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
aaruni96:
```suggestion

```

.github/workflows/tutorial_status.yml
aaruni96: We can later simplify the script to only care about the env var `API_KEY`. When testing locally, seed it with your Github personal access token. When working on the CI, seed it with `secrets.GITHUB_TOKEN`.

.github/workflows/test-contributer-script.yml
```diff
@@ -30,22 +30,78 @@ jobs:
...
+      uses: mikefarah/yq@v4
+      with:
+        cmd: |
+          test -f _data/people_list.yml
+          yq -e 'type == "!!seq"' _data/people_list.yml >/dev/null
+          echo "PEOPLE_COUNT_BEFORE=$(yq -r 'length' _data/people_list.yml)" >> $GITHUB_ENV
+
     - name: Run update-contributors script
       env:
         API_KEY: ${{secrets.API_KEY}}
```
aaruni96:
```suggestion
        API_KEY: ${{ secrets.GITHUB_TOKEN }}
```

.github/workflows/test-contributer-script.yml
```diff
@@ -30,22 +30,78 @@ jobs:
...
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
aaruni96:
```suggestion

```

.github/workflows/tutorial-last-modified-dates.yml
```diff
@@ -26,10 +26,58 @@ jobs:
           python -m pip install --upgrade pip
           python -m pip install -r etc/requirements.txt
 
-      - name: Run Tutorial Last Modified Date Script
+      - name: Update Last-Modified Dates for Tutorials
         env:
           API_KEY: ${{ secrets.API_KEY }}
```
aaruni96:
```suggestion
          API_KEY: ${{ secrets.GITHUB_TOKEN }}
```

.github/workflows/tutorial-last-modified-dates.yml
```diff
@@ -26,10 +26,58 @@ jobs:
...
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
aaruni96:
```suggestion

```

aaruni96 (commented):
Minor comments about the API_KEY, and a general comment about the test for the latest OSCAR release.

After this, should be ready.

.github/workflows/release-checklist.yml
HereAround: I made changes to the release CI tests as discussed on slack. Unless I am mistaken, this should mean this here is resolved. Please correct me if I am wrong.

HereAround:
> Minor comments about the API_KEY, and a general comment about the test for the latest OSCAR release.
> 
> After this, should be ready.

@aaruni96 I updated the Sync OSCAR Release Information CI job as discussed on slack and just merged your code suggestions. So I believe this is ready for another round of review, maybe even merge.

.github/workflows/release-checklist.yml
```diff
@@ -34,16 +35,60 @@ jobs:
...
+        run: |
+          cp -f _data/release.yml _data/old_release.yml
+          cat > _data/release.yml <<'YAML'
+          version: "0"
+          year: "1900"
+          month: "01"
+          day: "01"
+          date: "1900-01-01"
+          julia-min: "0.0"
+          YAML
```
fingolfin: Hmmm.. but why do you need to backup this file? You are working in a git repository, and can simply use git to restore the content, or detect if it was modified.

I also find the use of "sentinel" odd here. Maybe "reference version", or... just omit it?

That said, my original suggestion was to really *delete* the file. That way your step `Ensure _data/release.yml ` later on also makes more sense to me: it checks if the file was recreated by the script (as opposed to "the script did not thing").

```suggestion
      - name: Delete release.yml
        run: rm -f _data/release.yml
```

.github/workflows/release-checklist.yml
```diff
@@ -34,16 +35,60 @@ jobs:
...
+
+      - name: Sync OSCAR release information
         env:
           API_KEY: ${{ secrets.GITHUB_TOKEN }}
           TARGET_REPO: oscar-system/Oscar.jl
         run: |
-          python3 etc/update-latest-release.py
+          python3 etc/sync_oscar_release_information.py
+
+      - name: Ensure _data/release.yml exists
```
fingolfin: Maybe
```suggestion
      - name: Ensure _data/release.yml exists and is non-empty
```
to match what the step does (it uses `test -s`, not `test -e` or `test -f`).

That said, you could also just drop this step, as the next step also will fail if the file is missing?

.github/workflows/release-checklist.yml
```diff
@@ -34,16 +35,60 @@ jobs:
...
+          if dpkg --compare-versions "$new" gt "$old"; then
+            echo "changed=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "changed=false" >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: Cleanup temporary release backup
+        if: always()
+        run: |
+          rm -f _data/old_release.yml
```
fingolfin:
```suggestion
      - name: Determine whether _data/release.yml changed
        id: release
        run: |
          if git diff-index --quiet HEAD -- _data/release.yml; then
            echo "changed=false" >> "$GITHUB_OUTPUT"
          else
            echo "changed=true" >> "$GITHUB_OUTPUT"
          fi
```

.github/workflows/release-checklist.yml
```diff
@@ -34,16 +35,60 @@ jobs:
...
         run: |
-          python3 etc/update-latest-release.py
+          python3 etc/sync_oscar_release_information.py
+
+      - name: Ensure _data/release.yml exists
+        run: |
+          test -s _data/release.yml
+
+      - name: Validate _data/release.yml structure
+        uses: mikefarah/yq@v4
```
fingolfin: This action runs its commands in a docker container. Why do we need to do that? Isn't `yq` preinstalled on GitHub Actions anymore?

In fact I note that in the next step, you don't use `mikefarah/yq@v4`, but you *do* use `yq`... So can we try
```suggestion
```

.github/workflows/test-contributer-script.yml
```diff
@@ -30,22 +30,77 @@ jobs:
         python -m pip install --upgrade pip
         python -m pip install -r etc/requirements.txt
 
+    - name: Record current number of contributors (listed in people_list.yml)
+      uses: mikefarah/yq@v4
```
fingolfin: Same question as above: do we really need this?
```suggestion
```

.github/workflows/tutorial-last-modified-dates.yml
```diff
@@ -26,10 +26,49 @@ jobs:
...
           PYTHONUNBUFFERED: "1"
         run: |
-          python3 etc/update-dates.py
+          python3 etc/update_last_modified_dates_for_tutorials.py
+
+      - name: Ensure _data/examples_with_updated_last_modified.yml exists
+        run: |
+          test -s _data/examples_with_updated_last_modified.yml
+
+      - name: Show _data/examples_with_updated_last_modified.yml exists
```
fingolfin: Isn't this redundant? The `cat `will fail if the file does not exist?
```suggestion
      - name: Show _data/examples_with_updated_last_modified.yml
```

.github/workflows/tutorial_status.yml
```diff
@@ -26,10 +26,32 @@ jobs:
...
+          API_KEY: ${{ secrets.GITHUB_TOKEN }}
           PYTHONUNBUFFERED: "1"
         run: |
-          python3 etc/tutorial_status.py
+          python3 etc/update_tutorial_status.py
+
+      - name: Ensure _data/examples_status.yml exists
+        run: |
+          test -s _data/examples_status.yml
+
```
fingolfin: As above: What do we gain if the next step will fail anyway if the file is missing?
```suggestion
```

.github/workflows/release-checklist.yml
```diff
@@ -34,16 +35,60 @@ jobs:
...
+          cat > _data/release.yml <<'YAML'
+          version: "0"
+          year: "1900"
+          month: "01"
+          day: "01"
+          date: "1900-01-01"
+          julia-min: "0.0"
+          YAML
+
+      - name: Sync OSCAR release information
```
fingolfin: This comment is very nitpicky, but to me "fetch" or "query" seems more appropriate than "Sync". But feel free to ignore this comment :-)

.github/workflows/release-checklist.yml
fingolfin: Of course this does not reject the case where the version went "down". But that's a feature, IMHO: if that happens, something is really wrong, and we want to know.

fingolfin (commented):
Thanks a lot! I have a bunch of further comments. Sorry for being late to the party but I was ill.

HereAround:
> Thanks a lot! I have a bunch of further comments. Sorry for being late to the party but I was ill.

Thank you for the feedback. I believe you have good points here. Let me look into implementing the corresponding changes.

Brief feedback/rational:
1. Why not delete the release.yml? For this, I will need to modify the python script - it is set up to fail if the file is not present. I was avoiding this modification, in that I was hoping this PR would only make changes to the CI. But then, since both you and @aaruni96 brought this up, it would seem that modifying the python script is the cleaner solution. I will make this happen now. (For your curiosity: With the current change, the script would always thinks that the recent fetch from github is a new release, and therefore overwrite the release file.)
2. Regarding the checks for "is the file existing": I suppose you have a good point here point - displaying the content should indeed fail if the file in question is not present. I was just trying to be thorough in the CI: First check if the file exists and is non-empty, then display it. No deeper reason.

Let me go about these improvements...

.github/workflows/release-checklist.yml
aaruni96: The script  (which can also be improved, but we avoided trying to modify the script in this PR) expects `_data/release.yml` to exist at the beginning, to read `version` from it. The script then compares this version to the version fetched from github, and decides whether a new version has been released and to do something, or that no new release has happened, and to do nothing.

My suggestion was, since we expect (and check for) a proper yaml scheme at the end, but the script only needs `version` to work, to do something like


```suggestion
      - name: Rest release file to be empty
        run: |
          cat > _data/release.yml <<'YAML'
          version: "0"
          YAML
```

i.e, have the `release.yml` file be in a state that the script still executes, but fail the validator later on in the CI run, if the script hasn't rewritten it with "good" content (at least "well shaped" content).

.github/workflows/release-checklist.yml
aaruni96: This was originally a part of my review, until I realised, `test -s` checks for both existing, and non empty. The next step only fails if it doesn't exist, but passes if it exists and is empty, right ?

.github/workflows/release-checklist.yml
HereAround: I will improve this.

What is definitely true is that the script only needs the version, so @aaruni96 suggestion is sufficient/makes the CI cleaner. Will thus merge this change.

.github/workflows/release-checklist.yml
HereAround: Added the "and is non-empty".

Regarding the dropping - see the comment by @aaruni96 above.

.github/workflows/release-checklist.yml
HereAround: I am updating to indeed delete `_data/release.yml` (discussed this step on slack with @aaruni96 and we believe it is indeed a good improvement).

.github/workflows/tutorial-last-modified-dates.yml
HereAround: See the comment by @aaruni96 .

.github/workflows/tutorial_status.yml
HereAround: Similarly, see the comment by @aaruni96 .

.github/workflows/release-checklist.yml
HereAround: Yes, this seems nice.

.github/workflows/release-checklist.yml
HereAround: From what I read at https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md and https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md, indeed `yq` is available for ubuntu `24.04` (= latest). So yes, we can delete this. And I have done so.

.github/workflows/test-contributer-script.yml
HereAround: You are right, can be deleted. Have done this.

HereAround:
@fingolfin @aaruni96 I hope to have addressed all (or at least most) of the points raised. In particular:

- Delete the release.yml file (done by the python script), then write its content anew with the intel obtained from the github API.
- Removed "uses: mikefarah/yq@v4" in all workflows.
- Use `git diff` to detect a version change.
- I have not deleted the tests for "file exists and is non-empty", cf. https://github.com/oscar-system/oscar-website/pull/741#discussion_r2720768382.

With that, ready for another round of reviews. Thank you!

fingolfin (approved):

.github/workflows/release-checklist.yml
```diff
@@ -56,8 +82,8 @@ jobs:
           update_existing: true
           search_existing: open          
 
-      - name: Create PR updating latest release
-        if: steps.release.outputs.changed == 'true'
+      - name: Create pull request which updates _data/release.yml
+        if: github.event_name != 'pull_request' && !cancelled() && steps.release.outputs.changed == 'true'
```
```suggestion
        if: github.event_name != 'pull_request' && !cancelled()
```

The PR action only creates a PR if there are changes. We can skip checking for it ourselves. But this is a minor thing, and one could argue that being verbose in checks is preferable.

.github/workflows/tutorial-last-modified-dates.yml
```diff
@@ -26,10 +26,45 @@ jobs:
...
+          yq -e '
+            type == "!!seq" and length > 0 and
+            (
+              map(
+                (type == "!!map") and
+                has("author")     and (.author     | type == "!!str" and length > 0) and
+                has("branch")     and (.branch     | type == "!!str" and length > 0) and
+                has("date")       and (.date       | type == "!!str" and length > 0) and
+                has("filename")   and (.filename   | type == "!!str" and length > 0) and
+                has("language")   and (.language   | type == "!!str" and length > 0) and
```
aaruni96, line 58: We can skip the check for language, and later on, also actually remove language from our data files, because its not used anywhere.

```suggestion
```

etc/fetch_oscar_release_information.py
```diff
@@ -17,6 +17,13 @@
 DATAPATH = os.path.join(REPOPATH, "_data")
 RELEASEFILEPATH = os.path.join(DATAPATH, "release.yml")
 
+# Delete the release file should it exist
+try:
+    os.remove(RELEASEFILEPATH)
+    print(f"Deleted existing release file: {RELEASEFILEPATH}")
+except FileNotFoundError:
+    pass
```
aaruni96: We should always expect a file here. Its nice to wrap it in a try/except, but I would also like a warning here. Something like "RELEASEFILEPATH was not found. You may want to double check that everything works. Script will still continue..."

aaruni96 (approved):
A few very minor things, but otherwise, looks good!

etc/fetch_oscar_release_information.py
aaruni96:
```suggestion
    print(f"{RELEASEFILEPATH} was not found! This may indicate a weird bug of some sort. The script will still continue, but you should double check that everything works as expected.")
```

PR #742, opened 2026-01-23
Switch to new mailing list

fingolfin:
I plan to announce the new lists shortly, and then this PR should be merged soon-ish.

Note that it currently only mentions <devel@oscar-system.org> but not <users@oscar-system.org>. Which makes sense in so far as that the latter has had basically no traffic in the last few years, and we sent all announcements etc. to the former; but only a few to the latter. We may wish to change that in the future, or not (depending also on how successful we are in general... and what other communication platforms we can make available...)

Anyway, subscriptions for the two lists can be managed via

  https://oscar-system.org/lists/devel

and

  https://oscar-system.org/lists/users

These are redirects to the real sites (which right now are at uni-kl.de but will probably migrate this year to rptu.de domains...)

### DIFF
--- a/.github/release-checklist-template.md
+++ b/.github/release-checklist-template.md
@@ -8,7 +8,7 @@ labels: release-process
 
 A new OSCAR release **v{{ env.VERSION }}** was published on **{{ env.PUBLISHED_UTC }}** (UTC)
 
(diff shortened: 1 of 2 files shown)


HereAround (approved):
Looks good to me @fingolfin . As far as I am concerned, ready to be merged.

lgoettgens (approved):

PR #752, opened 2026-02-02
Fix and update CI

HereAround:
1. Not every entry in `_data/people_list.yml` has a github entry; we cannot determine these automatically for coauthors. Still, the CI tests assume so. This requires some changes.
2. For simpler debugging, print any entry in `_data/people_list.yml` that violates our criteria. This should help to debug in the future.

Cf. https://github.com/oscar-system/oscar-website/issues/753


aaruni96 (approved):
My `yq` knowledge is not too great, but this looks mostly plausible.

PR #755, opened 2026-02-02
Only create bot PRs if the workflow is a success

### LINKED ISSUE #753: Update contributor PR does not run our CI tests
E.g. https://github.com/oscar-system/oscar-website/pull/748.

Originally (this may change once we work on this), this PR suggested the following addition to `_data/people_list.yml`:

> - name: Igor Makhlin
>   email: 150450239+imkhln@users.noreply.github.com
>   github: imkhln
>   status: active
>   repos:
>   - oscar-system/Oscar.jl
> - name: Igor Makhlin
>   email: makhlin@artin.math.tu-berlin.de
>   status: active
>   comment: Co-author of commit af7cb7f6675714000774610a5ea8d7abaeed6728
>   repos:
>   - oscar-system/Oscar.jl

This change makes our CI test (update contributors) fail, because this currently requires that every entry in `_data/people_list.yml` has a `github` entry. Clearly, this is not the case for the coauthor added here. And indeed, this is our logic in how we add coauthors. It would be lovely if we could notice this directly in the PR for "Update contributors". I suggest the following:

1. Ideally, the update contributor script should recognize that the above entries belong to the same person, I suggest a fix to this in https://github.com/oscar-system/oscar-website/pull/754.
2. In the update contributor PRs - opened by our bot - the CI tests should run, and alert us to failures. A proposal for this is in https://github.com/oscar-system/oscar-website/pull/755.
3. The actual CI needs fixing. Not every entry has a guaranteed `github` entry. Also, it would be lovely to print any violating entry, as this will help with debugging in the future. I suggest this exact CI update/fix in https://github.com/oscar-system/oscar-website/pull/752.

cc @aaruni96

HereAround:
(~~Includes the fix in https://github.com/oscar-system/oscar-website/pull/752, as otherwise the CI will fail.~~ Already merged.)

Fixes https://github.com/oscar-system/oscar-website/issues/753.

### DIFF
--- a/.github/workflows/release-checklist.yml
+++ b/.github/workflows/release-checklist.yml
@@ -71,7 +71,7 @@ jobs:
           echo "published=$published" >> "$GITHUB_OUTPUT"
 
       - name: Create issue with release checklist
-        if: github.event_name != 'pull_request' && !cancelled() && steps.release.outputs.changed == 'true'
+        if: github.event_name != 'pull_request' && success() && steps.release.outputs.changed == 'true'
         uses: JasonEtco/create-an-issue@v2
         # marketplace action : https://github.com/marketplace/actions/create-an-issue
         env:
--- a/.github/workflows/release-checklist.yml
+++ b/.github/workflows/release-checklist.yml
@@ -83,10 +83,10 @@ jobs:
           search_existing: open          
 
       - name: Create pull request which updates _data/release.yml
-        if: github.event_name != 'pull_request' && !cancelled()
+        if: github.event_name != 'pull_request' && success()
         uses: peter-evans/create-pull-request@v7
         with:
-          token: ${{ secrets.PERSONAL_ACCESS_TOKEN }}
+
           commit-message: "Update latest OSCAR release to v${{ steps.release.outputs.version }}"
           title: "Update latest OSCAR release to v${{ steps.release.outputs.version }}"
           body: |
--- a/.github/workflows/release-checklist.yml
+++ b/.github/workflows/release-checklist.yml
@@ -86,6 +86,7 @@ jobs:
         if: github.event_name != 'pull_request' && !cancelled()
         uses: peter-evans/create-pull-request@v7
         with:
+          token: ${{ secrets.PERSONAL_ACCESS_TOKEN }}
           commit-message: "Update latest OSCAR release to v${{ steps.release.outputs.version }}"
           title: "Update latest OSCAR release to v${{ steps.release.outputs.version }}"
           body: |
--- a/.github/workflows/release-checklist.yml
+++ b/.github/workflows/release-checklist.yml
@@ -85,10 +85,9 @@ jobs:
           search_existing: open          
 
       - name: Create pull request which updates _data/release.yml
-        if: github.event_name != 'pull_request' && !cancelled()
+        if: github.event_name != 'pull_request' && success()
         uses: peter-evans/create-pull-request@v8
         with:
-
           commit-message: "Update latest OSCAR release to v${{ steps.release.outputs.version }}"
           title: "Update latest OSCAR release to v${{ steps.release.outputs.version }}"
           body: |
--- a/.github/workflows/test-contributer-script.yml
+++ b/.github/workflows/test-contributer-script.yml
@@ -103,6 +103,7 @@ jobs:
       if: github.event_name != 'pull_request' && !cancelled()
       uses: peter-evans/create-pull-request@v8
       with:
+        token: ${{ secrets.PERSONAL_ACCESS_TOKEN }}
         commit-message: "Update contributors on ${{ env.TODAY }}"
         branch: update-contributors
         delete-branch: true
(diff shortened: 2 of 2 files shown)


.github/workflows/release-checklist.yml
```diff
@@ -86,6 +88,7 @@ jobs:
         if: github.event_name != 'pull_request' && !cancelled()
         uses: peter-evans/create-pull-request@v7
         with:
+          token: ${{ secrets.PERSONAL_ACCESS_TOKEN }}
```
```suggestion
          token: ${{ secrets.GITHUB_TOKEN }}
```

.github/workflows/test-contributer-script.yml
```diff
@@ -104,6 +103,7 @@ jobs:
       if: github.event_name != 'pull_request' && !cancelled()
       uses: peter-evans/create-pull-request@v8
       with:
+        token: ${{ secrets.PERSONAL_ACCESS_TOKEN }}
```
aaruni96:
```suggestion
        token: ${{ secrets.GITHUB_TOKEN }}
```

aaruni96 (changes requested):
We should use `GITHUB_TOKEN` while running on CI. Other than that, looks good.

aaruni96:
Now that #752 is merged, this should maybe be rebased to `gh-pages` once before merge ?

HereAround:
> Now that #752 is merged, this should maybe be rebased to `gh-pages` once before merge ?

Just rebased.

.github/workflows/release-checklist.yml
HereAround: Look at https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/trigger-a-workflow. I quote:

> For example, the following workflow uses a personal access token (stored as a secret called MY_TOKEN) to add a label to an issue via GitHub CLI. Any workflows that run when a label is added will run once this step is performed.
> 
> on:
>   issues:
>     types:
>       - opened
> 
> jobs:
>   label_issue:
>     runs-on: ubuntu-latest
>     steps:
>       - env:
>           GH_TOKEN: ${{ secrets.MY_TOKEN }}
>           ISSUE_URL: ${{ github.event.issue.html_url }}
>         run: |
>           gh issue edit $ISSUE_URL --add-label "triage"
> 
> Conversely, the following workflow uses GITHUB_TOKEN to add a label to an issue. It will not trigger any workflows that run when a label is added.

So I believe we need a token different from `GITHUB_TOKEN` so that our workflows/CIs are run. And I found that we have one other token in the repository's secrets, namely the `PERSONAL_ACCESS_TOKEN`. Makes sense?

Alternatively, we can try first with the `GITHUB_TOKEN` to see if the CI runs in the bot PRs (contributor update for instance).

.github/workflows/test-contributer-script.yml
HereAround: See above.

.github/workflows/release-checklist.yml
HereAround: Related: I guess that by design, git sets the dependabot to triggers workflows/CIs.

HereAround: (Of course we could also try with a token specifically created and used only for these automated bot PRs.)

.github/workflows/release-checklist.yml
```diff
@@ -86,6 +86,7 @@ jobs:
         if: github.event_name != 'pull_request' && !cancelled()
```
```suggestion
        if: github.event_name != 'pull_request' && success()
```

.github/workflows/release-checklist.yml
```diff
@@ -86,6 +86,7 @@ jobs:
...
         uses: peter-evans/create-pull-request@v7
         with:
+          token: ${{ secrets.PERSONAL_ACCESS_TOKEN }}
```
```suggestion

```

.github/workflows/test-contributer-script.yml
```diff
@@ -103,6 +103,7 @@ jobs:
       if: github.event_name != 'pull_request' && !cancelled()
```
aaruni96:
```suggestion
      if: github.event_name != 'pull_request' && success()
```

.github/workflows/test-contributer-script.yml
```diff
@@ -103,6 +103,7 @@ jobs:
...
       uses: peter-evans/create-pull-request@v8
       with:
+        token: ${{ secrets.PERSONAL_ACCESS_TOKEN }}
```
aaruni96:
```suggestion

```

aaruni96 (changes requested):
We already test and validate the files we update as part of the workflow. Making a PR is only the final step. As such, we don't want to run CI on the PRs created by these workflows. The real problem is that the PRs were being created even when the validation step failed (see https://github.com/oscar-system/oscar-website/actions/runs/21553533212/job/62105942503#step:12:1 )

The suggested changes should fix that, and we have no need for adding tokens.

HereAround:
> We already test and validate the files we update as part of the workflow. Making a PR is only the final step. As such, we don't want to run CI on the PRs created by these workflows. The real problem is that the PRs were being created even when the validation step failed (see https://github.com/oscar-system/oscar-website/actions/runs/21553533212/job/62105942503#step:12:1 )
> 
> The suggested changes should fix that, and we have no need for adding tokens.

Alright. Changes applied accordingly. With this, ready for your review @aaruni96 .

aaruni96 (approved):
Thanks for the work on this!

PR #761, opened 2026-02-03
More CI fixes

### LINKED ISSUE #758: Release issue opens repeatedly
I recently closed an issue like https://github.com/oscar-system/oscar-website/issues/756. Our CI should not have opened this anew. We need a mechanism to tell if such an issue existed previously and was closed, and if so to not create the issue anew.

In my first attempt, the logic in the CI was to say "Was this release more than x days ago? Then do not open an issue." The variable x was set to say 5 days or alike.

Of course, I understand this is not perfect. And I believe this was overwritten at some point in an attempt to improve the workflow. Sadly, it seems the new logic does not prevent repeated opening of this issue. So will need to re-investigate.

HereAround:
Fixes https://github.com/oscar-system/oscar-website/issues/758.

### DIFF
--- a/.github/workflows/release-checklist.yml
+++ b/.github/workflows/release-checklist.yml
@@ -117,6 +117,7 @@ jobs:
             _Opened automatically by a scheduled workflow._
           branch: "bot/update-latest-release-v${{ steps.release.outputs.version }}"
           base: gh-pages
+          delete-branch: true
           labels: "release-process"
           author: "update_release[bot] <update_release[bot]@users.noreply.github.com>"
           committer: "update_release[bot] <update_release[bot]@users.noreply.github.com>"
--- a/.github/workflows/release-checklist.yml
+++ b/.github/workflows/release-checklist.yml
@@ -60,9 +60,9 @@ jobs:
       - name: Determine whether _data/release.yml changed
         id: release
         run: |
-          set -euo pipefail
+          set -uo pipefail
 
-          set +e
+
           git diff --quiet HEAD -- _data/release.yml
           ec=$?
           set -e
@@ -72,8 +72,8 @@ jobs:
           elif [ $ec -eq 1 ]; then
             echo "changed=true" >> "$GITHUB_OUTPUT"
           else
-            echo "git diff failed with exit code $ec" >&2
-            git status --porcelain=v1 >&2 || true
+            echo "git diff failed with exit code $ec"
+            git status --porcelain=v1 || true
             exit $ec
           fi
 
--- a/.github/workflows/release-checklist.yml
+++ b/.github/workflows/release-checklist.yml
@@ -115,7 +115,7 @@ jobs:
           filename: .github/release-checklist-template.md
           assignees: HereAround, aaruni96
           update_existing: true
-          search_existing: all
+          search_existing: open
 
       - name: Create pull request which updates _data/release.yml
         if: github.event_name != 'pull_request' && !cancelled()
--- a/.github/workflows/release-checklist.yml
+++ b/.github/workflows/release-checklist.yml
@@ -60,11 +60,23 @@ jobs:
       - name: Determine whether _data/release.yml changed
         id: release
         run: |
-          if git diff-index --quiet HEAD -- _data/release.yml; then
+          set -euo pipefail
+
+          set +e
+          git diff --quiet HEAD -- _data/release.yml
+          ec=$?
+          set -e
+
+          if [ $ec -eq 0 ]; then
             echo "changed=false" >> "$GITHUB_OUTPUT"
-          else
+          elif [ $ec -eq 1 ]; then
             echo "changed=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "git diff failed with exit code $ec" >&2
+            git status --porcelain=v1 >&2 || true
+            exit $ec
           fi
+
           version="$(yq -r '.version' _data/release.yml)"
           echo "version=$version" >> "$GITHUB_OUTPUT"
           published="$(yq -r '.date' _data/release.yml)"
--- a/.github/workflows/release-checklist.yml
+++ b/.github/workflows/release-checklist.yml
@@ -85,22 +85,22 @@ jobs:
       - name: Debug changed/version and diff
         if: always()
         run: |
-          echo "event_name: ${{ github.event_name }}"
-          echo "ref:       ${{ github.ref }}"
-          echo "sha:       ${{ github.sha }}"
-          echo "changed:   ${{ steps.release.outputs.changed }}"
-          echo "version:   ${{ steps.release.outputs.version }}"
+          echo "event_name:             ${{ github.event_name }}"
+          echo "ref:                    ${{ github.ref }}"
+          echo "sha:                    ${{ github.sha }}"
+          echo "new release?:           ${{ steps.release.outputs.changed }}"
+          echo "latest OSCAR version:   ${{ steps.release.outputs.version }}"
 
           echo ""
           echo "== git status =="
           git status --porcelain=v1
 
           echo ""
-          echo "== diff for _data/release.yml (if any) =="
+          echo "== git diff for _data/release.yml (if any) =="
           git --no-pager diff -- _data/release.yml || true
 
           echo ""
-          echo "== yq output (normalized) =="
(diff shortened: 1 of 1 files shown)


.github/workflows/release-checklist.yml
```diff
@@ -60,16 +60,49 @@ jobs:
...
+          git diff --quiet HEAD -- _data/release.yml
+          ec=$?
+          set -e
+
+          if [ $ec -eq 0 ]; then
             echo "changed=false" >> "$GITHUB_OUTPUT"
-          else
+          elif [ $ec -eq 1 ]; then
             echo "changed=true" >> "$GITHUB_OUTPUT"
+          else
```
aaruni96, line 72: Does `git diff` ever throw an exit code that is neither 0, nor 1?

If the answer is "no", then I vote to keep the previous version (maybe replace `diff-index` with `diff`). 

The old version is much easier to read, IMO (if \<something\> then false, otherwise true).

The new version is harder to parse:
- set -e
- actually set +e
- run a command and capture its output
- set -e again
- if exitcode equals 0 then false, elif equals 1 then true
- else (does this ever happen?) print some debug information and throw an error

Also, why are we redirecting output to stderr ?

.github/workflows/release-checklist.yml
```diff
@@ -60,16 +60,49 @@ jobs:
...
+            exit $ec
           fi
+
           version="$(yq -r '.version' _data/release.yml)"
           echo "version=$version" >> "$GITHUB_OUTPUT"
           published="$(yq -r '.date' _data/release.yml)"
           echo "published=$published" >> "$GITHUB_OUTPUT"
 
+      - name: Debug changed/version and diff
+        if: always()
```
aaruni96, line 84: Maybe we only want the debug output if we are trying to debug?


```suggestion
        if: runner.debug == '1'
```

.github/workflows/release-checklist.yml
```diff
@@ -82,10 +115,10 @@ jobs:
           filename: .github/release-checklist-template.md
           assignees: HereAround, aaruni96
           update_existing: true
-          search_existing: open          
+          search_existing: all
 
       - name: Create pull request which updates _data/release.yml
-        if: github.event_name != 'pull_request' && !cancelled()
+        if: github.event_name != 'pull_request' && !cancelled() && steps.release.outputs.changed == 'true'
```
aaruni96: In my opinion, this change is unnecessary.

The action already does not create a PR if there are no changes.

It doesn't hurt to keep it, but I would vote to get rid of it for (subjective to me) simplicity.

.github/workflows/release-checklist.yml
```diff
@@ -82,10 +115,10 @@ jobs:
           filename: .github/release-checklist-template.md
           assignees: HereAround, aaruni96
           update_existing: true
-          search_existing: open          
+          search_existing: all
```
aaruni96: I don't like this change... : 
At this point in the workflow, we should have already successfully decided whether or not there is a new release, and should only not open an issue if its already opened (if the issue is already opened, I would expect the action to update the issue, for example, if we change the template and manually run the action again, then it will keep the open issue and only update its contents).

If our workflow decides a new version is released, but actually there is an issue with the same name which has already been closed, there is a problem with the decision making before this point, and I would want to know...

aaruni96 (changes requested):

.github/workflows/release-checklist.yml
HereAround: I believe the output from `git diff` can be 0 (no changes), 1 (changes), and greater or equal to 2 which indicates an underlying failure (e.g. something like https://github.com/travis-ci/travis-ci/issues/6069).

Indeed, I believe that in our previous setup, the git command must have raised an error as I could verify that the diff was indeed empty. This thus triggered us to go into the else clause and proceed with `changed = true`: "Oh cool, there is a new OSCAR version".

The `set -e`, `set +e` ensures that the step enclosed does not fail the workflow, no matter what exit code we receive from git. Subsequently, we use this exist code (0, 1, or different from those two) to tell us whether there are no changes, there are changes (which by itself may have other origins than a new release) or there was an error from executing git.

Makes sense? If not, I am always happy to hear about better ideas. Thank you!

.github/workflows/release-checklist.yml
HereAround: I thought this is cheap and generally interesting to take a look at?

For instance, if we ever change the python script and suddenly it comes to conclude there is a new OSCAR release, I would like to see the diff to investigate.

.github/workflows/release-checklist.yml
HereAround: I was thinking about this scenario:

* We close the "release checklist issue".
* And we do not yet merge the "update release" PR (e.g. because we forget).

Then the current setup will create a new issue, just like last night. That is not desired, no?

(No strong feelings here, but at least the whitespaces we could delete. ;) )

.github/workflows/release-checklist.yml
HereAround: No strong feelings here either, am happy either way.

My rational was hoping to preserve some natural resources by exiting (hopefully) early. But I understand this here is rather minimal... (if existent)

In any case: I just noticed that the contributor workflow has no similar flag. For consistency, indeed we either drop this additional guard, or we add it to both workflows.

.github/workflows/release-checklist.yml
```diff
@@ -60,16 +60,49 @@ jobs:
       - name: Determine whether _data/release.yml changed
         id: release
         run: |
-          if git diff-index --quiet HEAD -- _data/release.yml; then
+          set -euo pipefail
```
```suggestion
          set -uo pipefail
```

.github/workflows/release-checklist.yml
```diff
@@ -60,16 +60,49 @@ jobs:
...
+
+          set +e
```
```suggestion

```

.github/workflows/release-checklist.yml
aaruni96: At the time I couldn't verify this, but indeed, `git diff` can throw exit codes other than 0 or 1. So this mostly makes sense. Still minor changes would make sense:

1. if we don't want to `set -e` initially, maybe we should just get rid of it, instead of setting it at the start, and then immediately unsetting it in the next line.

.github/workflows/release-checklist.yml
```diff
@@ -60,16 +60,49 @@ jobs:
...
+          ec=$?
+          set -e
+
+          if [ $ec -eq 0 ]; then
             echo "changed=false" >> "$GITHUB_OUTPUT"
-          else
+          elif [ $ec -eq 1 ]; then
             echo "changed=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "git diff failed with exit code $ec" >&2
```
```suggestion
            echo "git diff failed with exit code $ec"
```

.github/workflows/release-checklist.yml
```diff
@@ -60,16 +60,49 @@ jobs:
...
+            git status --porcelain=v1 >&2 || true
```
```suggestion
            git status --porcelain=v1 || true
```

.github/workflows/release-checklist.yml
aaruni96: I could go either way on this. Its very minor anyway.

.github/workflows/release-checklist.yml
aaruni96: In my opinion, we should not close the issue until we finish doing all the things the issue points at / is generated by.

And if after doing that, the workflow still thinks an issue should be (re)opened, I think I would like to know in that case.

In short, I think changing this to `all` might unintentionally hide symptoms of other problems that might creep in later.

We should get rid of the extra whitespace at the end, though.

.github/workflows/release-checklist.yml
aaruni96: > indeed we either drop this additional guard, or we add it to both workflows.

Pick whichever you like (my vote, dropping it in both places because its less work, but you can add it in both places if you like)

.github/workflows/release-checklist.yml
HereAround: Alright, leaving it at `search_existing: open` and removing the whitespaces.

.github/workflows/release-checklist.yml
HereAround: Done.

aaruni96 (approved):
LGTM. Thanks!

Issue #514, opened 2025-05-30
Restore CI tests for the website content to catch e.g. syntax errors in Markdown, and exercise the Python scripts


fingolfin:
Please look into that, @aaruni96

HereAround:
Potentially related/relevant: https://github.com/oscar-system/oscar-website/actions/runs/18531094076/job/52813773525

HereAround:
With https://github.com/oscar-system/oscar-website/pull/741, we should have a first set of presentable tests for the python scripts in place. Of course, improvements are possible. But I see Markdown errors as the next big target. Will try to get into that shortly.

HereAround:
https://github.com/oscar-system/oscar-website/pull/792 was just merged.

@fingolfin Are there further tests that you would like? Or specific improvements that you desire? Or are you satisfied and this issue can be closed?

fingolfin:
It would be even better if it used the exact same Python script as we use on the server. But I'm happy for now, so let's close it.

PR #764, opened 2026-02-06
Attempt towards a markdown checker workflow

HereAround:
This PR is a first step towards https://github.com/oscar-system/oscar-website/issues/514 by introducing automated Markdown checks. The idea is to create a minimal and non-disruptive starting point; the exact scope and strictness of Markdown checks can be refined further based on discussion. Right now, we do **not** enforce a strict Markdown style (yet), but focus on catching a small set of common and potentially problematic issues, such as missing language tags on fenced code blocks, trailing whitespace, missing final newlines.

A few comments/questions:
- Running `markdownlint` with its default configuration results in about 1800 errors. I restricted the configuration to a small set of rules that focus on correctness rather than style, cf. `.markdownlint-cli2.yaml`
- While fixing the remaining markdown errors, it became apparent that `README.md` seems (at least partially) outdated.  Maybe it is worth to explain the current workflows/tutorial tester practices there?
- Currently, `README.md` **is included** in the Markdown checks. However, this file is (to my knowledge) not rendered as part of the OSCAR website, but it *is* the main entry point for visitors to this GitHub repository. I am undecided whether it should be excluded from the markdown checks or not. Ideas?
- Similarly, a fenced code block without a language specification was found in `README.hyde.md`. I fixed it. However, I am undecided if this should be included in our test.

### DIFF
--- a/.github/workflows/markdownchecks.yml
+++ b/.github/workflows/markdownchecks.yml
@@ -26,4 +26,6 @@ jobs:
             !.github/**
             !etc/**
             !README.hyde.md
+            !LICENSE.hyde.md
+            !README.md
             !_sass/hyde.scss
--- a/.github/workflows/markdownchecks.yml
+++ b/.github/workflows/markdownchecks.yml
@@ -25,7 +25,7 @@ jobs:
             !_site/**
             !.github/**
             !etc/**
-            !README.hyde.md
+            !README*
             !LICENSE.hyde.md
             !README.md
             !_sass/hyde.scss
--- a/.github/workflows/markdownchecks.yml
+++ b/.github/workflows/markdownchecks.yml
@@ -1,0 +1,29 @@
+name: Markdown checks
+
+on:
+  pull_request:
+  push:
+    branches: [gh-pages]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  markdownchecks:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+
+      - name: Lint Markdown
+        uses: DavidAnson/markdownlint-cli2-action@v22
+        with:
+          globs: |
+            **/*.md
+            !vendor/**
+            !_site/**
+            !.github/**
+            !etc/**
+            !README.hyde.md
+            !_sass/hyde.scss
--- a/.github/workflows/markdownchecks.yml
+++ b/.github/workflows/markdownchecks.yml
@@ -21,11 +21,8 @@ jobs:
         with:
           globs: |
             **/*.md
-            !vendor/**
             !_site/**
             !.github/**
             !etc/**
             !README*
             !LICENSE.hyde.md
-            !README.md
-            !_sass/hyde.scss
--- a/.markdownlint-cli2.yaml
+++ b/.markdownlint-cli2.yaml
@@ -1,0 +1,31 @@
+# Markdown rules for CI
+#
+# We intentionally enable only a small set of rules that catch
+# likely mistakes or readability issues, not stylistic preferences.
+#
+# Rule documentation:
+#   https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md
+#
+config:
+  # Start from a clean slate: no rules enabled unless explicitly listed
+  default: false
+
+  # MD009 — no trailing spaces at end of lines
+  # (allows exactly two spaces for an intentional Markdown line break)
+  MD009:
+    br_spaces: 2
+    strict: false
+
+  # MD010 — disallow hard tabs (can render inconsistently)
+  MD010: true
+
+  # MD040 — fenced code blocks should specify a language
+  # (improves readability and syntax highlighting)
+  MD040: true
+
+  # MD047 — files should end with a single newline character
+  # (POSIX text file convention; avoids diff noise)
+  MD047: true
+
+  # Optional / currently disabled:
+  # MD031 — fenced code blocks should be surrounded by blank lines
--- a/.markdownlint-cli2.yaml
+++ b/.markdownlint-cli2.yaml
@@ -29,3 +29,4 @@ config:
 
   # Optional / currently disabled:
   # MD031 — fenced code blocks should be surrounded by blank lines
+  MD031: true


aaruni96:
I think it might make sense to split off 5a88d072 into another PR, so it makes it easier to see what the actual changes are which do the testing, and then separately, see the results of said testing.

benlorenz:
> I think it might make sense to split off [5a88d07](https://github.com/oscar-system/oscar-website/commit/5a88d072238a31aa92a40b8798af3cc8c7815856) into another PR, so it makes it easier to see what the actual changes are which do the testing, and then separately, see the results of said testing.

I would prefer not to change the non-markdown files at all (especially those that are sourced from other repositories).

HereAround:
> I think it might make sense to split off [5a88d07](https://github.com/oscar-system/oscar-website/commit/5a88d072238a31aa92a40b8798af3cc8c7815856) into another PR, so it makes it easier to see what the actual changes are which do the testing, and then separately, see the results of said testing.

Alright. Let me split it into two - one PR with the new action and failures that need fixing, a second with the proposed fixes.

HereAround:
> > I think it might make sense to split off [5a88d07](https://github.com/oscar-system/oscar-website/commit/5a88d072238a31aa92a40b8798af3cc8c7815856) into another PR, so it makes it easier to see what the actual changes are which do the testing, and then separately, see the results of said testing.
> 
> I would prefer not to change the non-markdown files at all (especially those that are sourced from other repositories).

Thank you. Aligns with my thoughts. Will adjust accordingly.

HereAround:
> > I think it might make sense to split off [5a88d07](https://github.com/oscar-system/oscar-website/commit/5a88d072238a31aa92a40b8798af3cc8c7815856) into another PR, so it makes it easier to see what the actual changes are which do the testing, and then separately, see the results of said testing.
> 
> Alright. Let me split it into two - one PR with the new action and failures that need fixing, a second with the proposed fixes.

Done. This PR now contains only the new action. The fixes to its exposed issues have been moved to https://github.com/oscar-system/oscar-website/pull/767.

aaruni96:
While the changes here look reasonable, the effects on the output (#767) are not fine (see my comment there: https://github.com/oscar-system/oscar-website/pull/767#issuecomment-3877164974 )

HereAround:
> While the changes here look reasonable, the effects on the output (#767) are not fine (see my comment there: [#767 (comment)](https://github.com/oscar-system/oscar-website/pull/767#issuecomment-3877164974) )

I suppose we want to first finish this PR and then turn to the fixes. As such, @aaruni96 can you take a closer look and tell me if you think more files/folders should be excluded from the tests? Thank you!

.github/workflows/markdownchecks.yml
```diff
@@ -0,0 +1,31 @@
...
+          globs: |
+            **/*.md
+            !vendor/**
+            !_site/**
+            !.github/**
+            !etc/**
+            !README.hyde.md
+            !LICENSE.hyde.md
+            !README.md
+            !_sass/hyde.scss
```
aaruni96:
```suggestion
```

This should not need to be manually excluded, since its not a md file to begin with.

.markdownlint-cli2.yaml
```diff
@@ -0,0 +1,31 @@
...
+  # MD040 — fenced code blocks should specify a language
+  # (improves readability and syntax highlighting)
+  MD040: true
+
+  # MD047 — files should end with a single newline character
+  # (POSIX text file convention; avoids diff noise)
+  MD047: true
+
+  # Optional / currently disabled:
+  # MD031 — fenced code blocks should be surrounded by blank lines
```
aaruni96, line 31:
```suggestion
  # MD031 — fenced code blocks should be surrounded by blank lines
  MD031: true
```

My personal preference is to have blank lines around code fences, it makes it somewhat cleaner to read. But this suggestion could also be ignored. (This flags 43 new errors, but automated fixing remains an option)

.github/workflows/markdownchecks.yml
```diff
@@ -0,0 +1,31 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+
+      - name: Lint Markdown
+        uses: DavidAnson/markdownlint-cli2-action@v22
+        with:
+          globs: |
+            **/*.md
+            !vendor/**
```
aaruni96:
```suggestion
```

This directory does not exist (and therefore need not be excluded?)

.github/workflows/markdownchecks.yml
```diff
@@ -0,0 +1,31 @@
...
+            !_site/**
+            !.github/**
+            !etc/**
+            !README.hyde.md
```
aaruni96:
```suggestion
            !README*
```

This will exclude all the readmes (currently two) with a single line of config

.github/workflows/markdownchecks.yml
```diff
@@ -0,0 +1,31 @@
...
+            !LICENSE.hyde.md
+            !README.md
```
aaruni96:
```suggestion
```

Since line 28 would address this

aaruni96 (changes requested):
Only minor changes, otherwise looks okay.

.markdownlint-cli2.yaml
aaruni96: If this change is applied, #768 will need to be updated as well.

.github/workflows/markdownchecks.yml
HereAround: My bad.

HereAround:
Thank you @aaruni96 . Just accepted all your suggestions. Please take another look.

aaruni96:
I think there was some problem (possibly Github getting confused about what to do with "empty" code suggestions), but lines 24, 30, and 31 still have the old data.

HereAround:
> I think there was some problem (possibly Github getting confused about what to do with "empty" code suggestions), but lines 24, 30, and 31 still have the old data.

Indeed, that is unusual; I added **all** your individual suggestions to a batch, then applied the changes. Still, it would seem that some suggestions were lost.

I have just pushed them manually. Please double check that I did not miss anything. Thank you @aaruni96 !

aaruni96:
Updated #768 with the changes from the latest config. Looks good, at least to my eyes.

@fingolfin can you also have a brief look ?

aaruni96 (approved):
Looks good to me, but I would wait for merge until Max chimes in.

HereAround:
From communication in slack with @fingolfin I understand, that this was not desired (as step towards https://github.com/oscar-system/oscar-website/issues/514). In private conversation with @aaruni96 I also heard disfavor to this action. As such, will close this now (not desired) and start fresh in a separate PR.

PR #765, opened 2026-02-09
Migrate hosting to newer server

fingolfin:
Our university will ask everyone to migrate to a new hosting environment and will decomission the old one. Let's stay ahead of the curve and migrate now (the 'new' server has already been in use for quite some time anyway, so it is already well-tested).

This updates our internal documentation to provide the new server hostname for SSH access. Also update some of our instructions as to what software is required (we now rely on pip, so it must be installed).


aaruni96 (approved):
LGTM, thanks!

PR #774, opened 2026-02-13
Ak96/debug release check

aaruni96:
Manually setting `+e` appears to be a requirement. (I thought that's the default, perhaps I'm mistaken).

Adding `-x` also seems like a good idea, because it causes the script to print the command its going to execute before it executes it, and thus helps in debugging if/when it fails.


HereAround (approved):
LGTM. Thank you!

PR #785, opened 2026-02-23
Filter out bot Codex

HereAround:


aaruni96:
I would start a list of "known_bots" (outside of github bots, because those can be checked with `[bot]` in the name), and have the full email and exact string matching. Otherwise, for example, a github user named "acodex" would be excluded by the check....

HereAround:
> I would start a list of "known_bots" (outside of github bots, because those can be checked with `[bot]` in the name), and have the full email and exact string matching. Otherwise, for example, a github user named "acodex" would be excluded by the check....

Good point. Please take a look at the updated proposal @aaruni96 .

aaruni96 (approved):
Looks plausible

PR #792, opened 2026-03-04
Attempt towards better website CI

HereAround:


fingolfin (approved):
Great, seems it detects the syntax error just fine

Issue #805, opened 2026-03-13
Add link back to OSCAR homepage from newsletters


fingolfin:
Nice job on setting https://www.oscar-system.org/newsletters/1.html and https://www.oscar-system.org/newsletters/2.html up. One minor request: those pages should have a link back to the main page, e.g. to https://www.oscar-system.org/newsletter/

Or perhaps the logo at the top could be clickable and go back to https://www.oscar-system.org/

HereAround:
With https://github.com/oscar-system/oscar-website/pull/809 merged, is there more that needs to be done @fingolfin ?

Issue #681, opened 2025-11-05
Add custom newsletter archive


fingolfin:
The first OSCAR newsletter was just delivered (CC @gfourier), yay.

But the the mailing list archive is a bit drab, see <https://lists.uni-kl.de/oscar/arc/news> (nevertheless, for now I added a link to it on <https://www.oscar-system.org/contact-and-support/>).

And unfortunately it does not show the images and styling (see <https://lists.uni-kl.de/oscar/arc/news/2025-11/msg00000.html>).

So perhaps we could add a new section on the website which archives the actual HTML versions of the newsletter, including styling). It wouldn't have to show up in the main navigation either...

Idea:
- add <https://www.oscar-system.org/newsletter/> with brief general information about the newsletter and how to sign up (BTW I think one can sign up by sending an email instead of using the webinterface, this process could then also be described)
- add <https://www.oscar-system.org/newsletter/archive/> with the past newsletters

That could all be fairly minimalistic.

fingolfin:
Another benefit from such an "online" version of the newsletter is that one can add links to the newsletter mails to it: "click here to view the online version in your web browser" or so -- in particular for people who have a mail client that does not render HTML, or who have disabled this; or who have (like me) disabled loading of images by default to prevent tracking.

HereAround:
I just took steps towards this, cf. https://github.com/oscar-system/oscar-website/pull/802. @gfourier seemed happy. Still, there might be room for improvement. @fingolfin what is your take? Feature requests? Modifications? Or can we close this issue?

HereAround:
With https://github.com/oscar-system/oscar-website/pull/809 merged, is there more that needs to be done @fingolfin ?

fingolfin:
Looks great, thanks!

PR #809, opened 2026-03-16
Make newsletter better

### LINKED ISSUE #806: Clarify subscription instruction on newsletter page
On https://www.oscar-system.org/newsletter/ it says:

> ## Subscribe
> Follow the instructions at https://lists.uni-kl.de/oscar/info/news.

But that page does not really have instructions one can follow. One can of course "guess" that one has to click "Subscribe" 

I think we should write a little bit more there. Like: "click on subscribe, then..". Perhaps borrow some from <https://lists.uni-kl.de/oscar/help/user>

We could also directly link to <https://lists.uni-kl.de/oscar/subscribe/news>

Finally I think it is also possible to subscribe by sending an email (this is also described at < https://lists.uni-kl.de/oscar/help/user>)

HereAround:
Should close https://github.com/oscar-system/oscar-website/issues/806, https://github.com/oscar-system/oscar-website/issues/805 and possibly also https://github.com/oscar-system/oscar-website/issues/681.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,7 @@
 - [How to contribute](#how-to-contribute)
 - [How to test the website locally](#how-to-test-the-website-locally)
 - [How to contribute a tutorial notebook](#how-to-contribute-a-tutorial-notebook)
+- [How to contribute a newsletter](#how-to-contribute-a-newsletter)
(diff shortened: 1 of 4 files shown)


HereAround:
@fingolfin I assume that this (minor) change is generally good and will thus merge it now. As per usual, am more than happy to improve/modify further.

PR #814, opened 2026-03-20
Add systemd timer to rebuild website every day

### LINKED ISSUE #797: Add cronjob to server to run update_tutorials.py regularly
Also update README.server.

cc @aaruni96

aaruni96:
@fingolfin can you have a quick look to see that this looks sane ?

### DIFF
--- a/etc/README.server.md
+++ b/etc/README.server.md
@@ -45,6 +45,13 @@ must be entered in the GitHub settings at
 <https://github.com/oscar-system/oscar-website/settings/hooks>.
 
 
+## Automatic daily rebuilds
+
+We also rebuild the website every day at around 3 AM. This fetches the latest
+testing status of the Tutorials by running the script `update_tutorials.py`. We
+do this via systemd timers, and the `oscar-website.timer` file in the `etc`
+directory.
+
 ## Github API
 
 The Python scripts in `etc/` use the Github API, and require an API key for
@@ -64,14 +71,14 @@ on the webserver.
 
 If updates stop working, a good first place to look at is this output of this:
 
-    systemctl --user status oscar-website.service oscar-website.path
+    systemctl --user status oscar-website.service oscar-website.path oscar-website.timer
 
 This prints a log with extra info. However, it might also say "service not
-found". In that case, make sure that `oscar-website.service` and
-`oscar-website.path` are installed and enabled:
+found". In that case, make sure that `oscar-website.service`,
+`oscar-website.path`, and `oscar-website.timer` are installed and enabled:
 
     cp ~/data/oscar-website/etc/oscar-website.* ~/.config/systemd/user
-    systemctl --user enable oscar-website.service oscar-website.path
+    systemctl --user enable oscar-website.service oscar-website.path oscar-website.timer
 
 Also helpful is to study the log for the relevant systemd units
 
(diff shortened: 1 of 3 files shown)


etc/README.server.md
```diff
@@ -139,8 +139,8 @@ Next install and activate the systemd units:
 
     mkdir -p ~/.config/systemd/user/
     cp ~/data/oscar-website/etc/oscar-website.* ~/.config/systemd/user/
-    systemctl --user enable oscar-website.service oscar-website.path
-    systemctl --user start oscar-website.service oscar-website.path
+    systemctl --user enable oscar-website.service oscar-website.path oscar-website.timer
+    systemctl --user start oscar-website.service oscar-website.path oscar-website.timer
```
fingolfin, line 150: There are multiple other places in this README that should be adjusted. Best to search for `oscar-website.service` and also `oscar-website.path` to find them

etc/oscar-website.timer
```diff
@@ -0,0 +1,8 @@
+[Unit]
+Description=Daily rebuilding of the website
+
+[Timer]
+OnCalendar=*-*-* 3:00:00    # every day at 3 AM localtime
```
fingolfin, line 5: never ever set cron jobs on a sharp time, that leads to load spikes. I used `rand(1:59)` twice to get
```suggestion
OnCalendar=*-*-* 3:41:03    # every day at 3:41 AM localtime
```

etc/oscar-website.timer
```diff
@@ -0,0 +1,8 @@
+[Unit]
+Description=Daily rebuilding of the website
```
fingolfin, line 2: I have no idea how systemd timers work. How does systemd know what do to trigger rebuilding the website? Is this implicit based on the filenames?

For `oscar-website.path` there is an explicit `Unit=oscar-website.service` in there to indicate which service to trigger. This is not needed here?

In other words: did you test locally that this works as intended?

etc/oscar-website.timer
aaruni96: Should we then use the inbuilt systemd randomiser?

The definition supports a `RandomizedDelaySec` which delays the start time by some amount of seconds chosen between 0 and `RandomizedDelaySec` uniformly.

etc/oscar-website.timer
aaruni96: One can explicitly specifiy `unit=`, but by default `<foo>.timer` triggers `<foo>.service`, and (quoting from the manpage)

>It is recommended that the unit name that is activated and the unit name of the timer unit are named identically, except for the suffix.

aaruni96: > In other words: did you test locally that this works as intended?

Not this timer specifically, but yes, I have other timers which run successfully on my system (which I used as a template for this one)

etc/oscar-website.timer
fingolfin: sounds good to me

etc/oscar-website.timer
fingolfin: OK makes sense, thanks!

Does a similar default for the unit to be triggered also hold for `.path` ? Then perhaps that one should also be adjusted?

fingolfin (dismissed):
Up to the randomization bit this seems ready to me!

etc/oscar-website.timer
aaruni96: Yes, a similar default, and recommendation exists for `.path`.

aaruni96:
I was apparently a bit overzealous when I set up branch protection rules (in principle it sounds like a great idea that newer commits should invalidate old approvals). I will force merge this now.

PR #832, opened 2026-05-04
Improve instructions for tutorials

HereAround:
Following suggestions by @JohnAAbbott , I propose to restructure the wording and instructions on the tutorial page.

### DIFF
--- a/_includes/tutorial.html
+++ b/_includes/tutorial.html
@@ -26,13 +26,10 @@ If you have not yet installed OSCAR, follow our <a href="{{site.baseurl}}/instal
 
 <h3>Running Tutorials Locally</h3>
 
-To download and run a tutorial on your machine:
-
 <ol>
-  <li>In the list of available tutorials below, open the tutorial of interest in either nbviewer or Github by clicking the respective link on each tutorial.</li>
-  <li>Right-click (or Ctrl-click on macOS) the download icon at the top-right corner.</li>
-  <li>Select <em>Save As...</em> and save the file to your computer.</li>
-  <li>Open Jupyter, load the downloaded notebook, and run it using the Julia kernel.</li>
+  <li>Download the notebook of interest from the list below.</li>
+  <li>Launch Jupyter on your machine (see <a href="https://jupyter.org/">Jupyter documentation</a>).</li>
+  <li>Open the downloaded notebook, and execute it using a Julia kernel.</li>
 </ol>
 
 ---
--- a/_includes/tutorial.html
+++ b/_includes/tutorial.html
@@ -29,7 +29,7 @@ If you have not yet installed OSCAR, follow our <a href="{{site.baseurl}}/instal
 To download and run a tutorial on your machine:
 
 <ol>
-  <li>In the list of available tutorials below, open the tutorial of interest in <a href="https://nbviewer.jupyter.org/"><em>nbviewer</em></a> (or view it directly on <a href="https://github.com/">GitHub</a>).</li>
+  <li>In the list of available tutorials below, open the tutorial of interest in either nbviewer or Github by clicking the respective link on each tutorial.</li>
   <li>Right-click (or Ctrl-click on macOS) the download icon at the top-right corner.</li>
   <li>Select <em>Save As...</em> and save the file to your computer.</li>
   <li>Open Jupyter, load the downloaded notebook, and run it using the Julia kernel.</li>
--- a/_includes/tutorial.html
+++ b/_includes/tutorial.html
@@ -2,7 +2,7 @@
 
 <p>
   Explore <a href="https://jupyter.org/">Jupyter notebooks</a> demonstrating OSCAR’s capabilities.
-  The tutorials are designed for and tested with the latest stable release <strong>OSCAR v{{ site.data.release.version }}</strong>.
+  The tutorials are designed for, and tested with, the latest stable release <strong>OSCAR v{{ site.data.release.version }}</strong>.
 </p>
 
 ---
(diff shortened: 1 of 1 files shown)


_includes/tutorial.html
```diff
@@ -2,36 +2,42 @@
...
+---
+
+<h3>Running Tutorials Locally</h3>
+
+To download and run a tutorial on your machine:
 
 <ol>
-  <li>Open the tutorial in <a href="https://nbviewer.jupyter.org/"><em>nbviewer</em></a> (or view it directly on <a href="https://github.com/">GitHub</a>).</li>
-  <li>Right-click (or Ctrl-click on macOS) the download icon at the top-right.</li>
+  <li>In the list of available tutorials below, open the tutorial of interest in <a href="https://nbviewer.jupyter.org/"><em>nbviewer</em></a> (or view it directly on <a href="https://github.com/">GitHub</a>).</li>
```
aaruni96: I would possibly get rid of links to nbviewer and github here. The link serves no purpose other than to possibly confuse an over-eager user. ( Im an inexperienced but eager user, I'm reading the tutorials page, it says "open the tutorial of interest in <link>". I clicked on link! Now I'm lost.... )

Maybe something like 



```suggestion
  <li>In the list of available tutorials below, open the tutorial of interest in either nbviewer or Github by clicking the respective link on each tutorial.</li>
```

_includes/tutorial.html
```diff
@@ -2,36 +2,42 @@
...
+  <li>Right-click (or Ctrl-click on macOS) the download icon at the top-right corner.</li>
```
aaruni96: This step assumes that the user picked nbviewer to download the tutorial. We should either document both methods, or get rid of one of the options above ?

(My vote is to get rid of nbviewer. As far as I can tell, its functionally equivalent to the Github page. And we probably want our users to become familiar with Github anyway, to report issues and hopefully eventually to contribute PRs)

aaruni96 (changes requested):
Minor comments.

_includes/tutorial.html
HereAround: If you don't mind, I would like to discuss/address "keeping support for nbviewer links" in a separate issue/PR.

My rational for this approach is two-fold:
1. I can see good reasons to keep supporting both github and nbviewer. Would like to hear from more people to gain a better understanding whether or not we want to remove the nbviewer links.
2. In this PR, I only wanted to address wording changes (presumably a low target, and quick to implement). However, getting rid of the nbviewer links takes us to css and alike, so clearly beyond what I intended the scope of this PR to be.

Hope you are ok with this @aaruni96? I will open an issue, so we do not forget about this matter. We could try to discuss/address this in full glory in the coding sprint later this week.

HereAround: Issue is here: https://github.com/oscar-system/oscar-website/issues/834

aaruni96: From my point of view, for the time we don't get rid of nbviewer (and maybe we ultimately vote to keep it in), we should still document what to do in case the user decides to click the link saying "github" instead of "nbviewer": 

If they click github, then they can't "right click the download icon at the top right corner"... (They must left click the "download raw file" icon)

HereAround: I was just thinking about a suitable wording choice for this... ;) Give me a minute or two.

_includes/tutorial.html
```diff
@@ -2,36 +2,42 @@
...
+
+<h3>Running Tutorials Locally</h3>
+
+To download and run a tutorial on your machine:
 
 <ol>
-  <li>Open the tutorial in <a href="https://nbviewer.jupyter.org/"><em>nbviewer</em></a> (or view it directly on <a href="https://github.com/">GitHub</a>).</li>
-  <li>Right-click (or Ctrl-click on macOS) the download icon at the top-right.</li>
+  <li>In the list of available tutorials below, open the tutorial of interest in either nbviewer or Github by clicking the respective link on each tutorial.</li>
+  <li>Right-click (or Ctrl-click on macOS) the download icon at the top-right corner.</li>
```
HereAround:
```suggestion
  <li>
  Download the notebook:
  in <strong>nbviewer</strong>, left-click the “Download notebook” button in the browser’s top-right corner;
  in <strong>GitHub</strong>, click the button near the file’s top-right edge (shows “Download raw file” on hover).
</li>
```

HereAround: Maybe like this?

HereAround: In particular, I don't think this requires "right clicking" (any more?). Rather, I see a button that I would left click.

HereAround:
@aaruni96 As discussed yesterday on slack, I have updated the instructions. They are now more abstract and should carry easily to the future (no matter what GUI changes, OS etc.). With that, this PR is ready for another round of review.

_includes/tutorial.html
```diff
@@ -2,36 +2,39 @@
 
 <p>
   Explore <a href="https://jupyter.org/">Jupyter notebooks</a> demonstrating OSCAR’s capabilities.
-  These tutorials are for <strong>OSCAR v{{ site.data.release.version }}</strong>,
-  the latest stable release. To avoid compatibility issues, make sure you're using this version
-  (<a href="{{ site.baseurl }}/upgrade">upgrade instructions</a>).
+  The tutorials are designed for and tested with the latest stable release <strong>OSCAR v{{ site.data.release.version }}</strong>.
```
JohnAAbbott: Minor punctuation niggle: put _and tested with_ between commas.

JohnAAbbott (approved):
Apart from one minor niggle, it seems fine to me now.

_includes/tutorial.html
HereAround:
```suggestion
  The tutorials are designed for, and tested with, the latest stable release <strong>OSCAR v{{ site.data.release.version }}</strong>.
```

HereAround: Done.

aaruni96 (approved):
LGTM, thanks!

PR #836, opened 2026-05-06
Uniform date format across the OSCAR website

HereAround:
As pointed out by @JohnAAbbott we are using different date formats across the OSCAR website. This PR aims to define a uniform start in config.yml and then apply this date standard across the entire OSCAR website.


JohnAAbbott:
**Quick summary:**  after a swift discussion we settled on  the format `00 Abc 0000`, namely 2 digits, a space, 3 letters (first is capital, the others lowercase(?)), a space, and 4 digits (for the year... we'll have a problem in 7974 years).  The format has the advantage that is it easy and unambiguous for a person to read, and has constant width.

credits/contributors.md
JohnAAbbott: Maybe change the date format in line 79 of `contributors.md`?  Or can one not use `site.data.config...` here?

index.md
JohnAAbbott, line 32: The format looks like ISO 8601 to me -- intentional?  `index.md` line 32

JohnAAbbott (approved):
I made some "reasonable guesses" about the syntax/semantics.  It seems good to me (with 2 explicit comments)

index.md
aaruni96: This value is only used for comparisons and internal logic, never printed. So, I would say, this is intentional.

aaruni96 (dismissed):
LGTM Thanks!

aaruni96 (approved):

PR #841, opened 2026-05-19
Additional date format adjustment and wording changes in release checklist template

HereAround:
This PR formats the github variable `published` in the release checklist workflow, so that in the release checklist issue, the date is formatted as uniformly applied across the website. As such, the email send out to inform about the latest release will employ the same date format as employed on the website.

I added a printout in the release checklist workflow to verify that the date has been formatted correctly, see https://github.com/oscar-system/oscar-website/actions/runs/26087062488/job/76702741539?pr=841 line 43.

### DIFF
--- a/.github/workflows/release-checklist.yml
+++ b/.github/workflows/release-checklist.yml
@@ -79,7 +79,7 @@ jobs:
           version="$(yq -r '.version' _data/release.yml)"
           echo "version=$version" >> "$GITHUB_OUTPUT"
           published="$(yq -r '.date' _data/release.yml)"
-          published="$(date -d "$published" '+%-d %b %Y')"
+          published="$(date -d "$published" '+%d %b %Y')"
           echo "published=$published" >> "$GITHUB_OUTPUT"
 
       - name: Debug changed/version and diff
--- a/.github/workflows/release-checklist.yml
+++ b/.github/workflows/release-checklist.yml
@@ -79,6 +79,7 @@ jobs:
           version="$(yq -r '.version' _data/release.yml)"
           echo "version=$version" >> "$GITHUB_OUTPUT"
           published="$(yq -r '.date' _data/release.yml)"
+          published="$(date -d "$published" '+%-d %b %Y')"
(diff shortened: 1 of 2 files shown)


.github/workflows/release-checklist.yml
```diff
@@ -79,6 +79,7 @@ jobs:
           version="$(yq -r '.version' _data/release.yml)"
           echo "version=$version" >> "$GITHUB_OUTPUT"
           published="$(yq -r '.date' _data/release.yml)"
+          published="$(date -d "$published" '+%-d %b %Y')"
```
aaruni96: `%-d` looks a bit strange to me. I see that it removes the leading 0 from the date if the date is between 0 and 9. Is this deliberate?

(One can argue whether keeping the leading 0 is better, or deleting it is better. I am not sure which one is better. I just want to make sure that this behaviour is deliberate and what is intended).

aaruni96 (dismissed):

.github/workflows/release-checklist.yml
HereAround: No this is not desired.

HereAround:
```suggestion
          published="$(date -d "$published" '+%d %b %Y')"
```

HereAround: Thank you for pointing this out.

HereAround: I presume this should work in keeping a leading 0. It matches `full_date_format: "%d %b %Y"` in our `config.yml`.

HereAround:
Just pushed a few minor wording changes for the template, so that we include links to the installation and upgrade instructions on the website - as suggested by @JohnAAbbott .

aaruni96 (approved):
LGMT thanks!

JohnAAbbott:
> Just pushed a few minor wording changes for the template, so that we include links to the installation and upgrade instructions on the website - as suggested by @JohnAAbbott .

Blitzschnell implementiert -- danke!  :-)

PR #844, opened 2026-05-20
Better onboarding

### LINKED ISSUE #843: Advertise community channels more
I talked to somebody on the M2 workshop who was struggling with the OSCAR docu and didn't know he could just post a question on the slack channel.  

I think it would be nice if we can advertise it (or whatever channel we end up with) more, e.g., a sentence right above "Upcoming Events" on the [fontpage](https://www.oscar-system.org/).

HereAround:
Aims to close https://github.com/oscar-system/oscar-website/issues/843.

Ping @YueRen


YueRen (approved):
Perfect!

PR #849, opened 2026-05-22
Add more contributor information

HereAround:
Our script might reorder the entries subsequently, could not do this yet.

Easiest is probably to merge this PR, wait for the follow-up PR by the bot and then merge this one. (Also, in case the keys need reordering.)


aaruni96 (approved):
Presumably all new intel is accurate.

Thanks for the detective work @HereAround !

PR #866, opened 2026-07-01
Fix minor irregularities discovered in the latest update-contributors action

aaruni96:
- Add Theo to known people
- Add chatGPT to known bots
- Fix summary string forming and printing

cc @HereAround

### DIFF
--- a/etc/update_contributors.py
+++ b/etc/update_contributors.py
@@ -46,8 +46,19 @@ REPO_LIST = (
 )
 
 # Known bot identities that we want to exclude, even if they don't have [bot] in name.
-KNOWN_BOT_EMAILS = {"codex@openai.com", "noreply@anthropic.com", "copilot@github.com"}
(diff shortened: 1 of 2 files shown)


etc/update_contributors.py
```diff
@@ -175,6 +186,12 @@ def find_coauthor_commit(name: str, email: str, repos: list[str]) -> str:
             m = HASH_RE.search(line)
             if m:
                 return m.group(0)
+    print("Warning! No commit has found for :")
```
HereAround:
```suggestion
    print("Warning! No commit hash found for :")
```

aaruni96: Thanks for the spell check!

HereAround (approved):
Thank you @aaruni96 . LGTM!

HereAround:
The summary string in the contributor script seems find again, looking at the output of the CI run above.

PR #883, opened 2026-08-02
Remove blog.html and fix broken links

HereAround:
As far as I see, we are not using blog.html (any more). I suggest to delete it.

I am also adding fixes for broken links in this PR. @ThomasBreuer this includes two links on your supplementary material for the Cogent Winter School. Could you please verify if this change is desired and points to the correct files? (See also https://github.com/oscar-system/oscar-website/issues/884)


aaruni96 (changes requested):
I agree with @ThomasBreuer , it is better to have files here, rather than link to external sources.

Regarding `blog.html`, it looks like around 6 years ago, it was moved to news instead. If we want to be super considerate, we could add a redirect for the old blog page to the new news page. And since we don't have a new news page, perhaps it could point to the newsletter page ? Or we simply say 6 years was enough time for people to forget about blog.

aaruni96 (approved):

PR #885, opened 2026-08-02
Improve accessibility and site metadata

HereAround:


_includes/tutorial.html
```diff
@@ -44,16 +44,16 @@ <h3>Available Tutorials</h3>
   {% assign topic_tutorial = site.data.tutorials | where: "thumbnail", component | first %}
   {% assign component_url = "/tutorials/" | append: component_name | append: "/" %}
   <div class="tutorial">
-    <a style="display: block; text-align: center;"
+    <a style="display: block; text-align: center;{% if page.url == component_url %} font-weight: bold;{% endif %}"
```
aaruni96: Why do we make this change? If I understand correctly, functionally it makes no change, but it adds yet another if clause and makes the line slightly less readable ?

_includes/sidebar.html
```diff
@@ -36,7 +36,7 @@
             <ul class="submenu">
               {% for subitem in item.submenu %}
                 <li class="{% if page.url contains subitem.url %}active{% endif %}">
-                  <a href="{% if subitem.url contains 'http' %}{{ subitem.url }}{% else %}{{ site.baseurl }}{{ subitem.url }}{% endif %}" {% if subitem.url contains 'http' %}target="_blank"{% endif %}>
+                  <a href="{% if subitem.url contains 'http' %}{{ subitem.url }}{% else %}{{ site.baseurl }}{{ subitem.url }}{% endif %}" {% if subitem.url contains 'http' %}target="_blank" rel="noopener noreferrer"{% endif %}>
```
aaruni96, line 39: What do these keywords do?

_includes/tutorial.html
```diff
@@ -44,16 +44,16 @@ <h3>Available Tutorials</h3>
...
           href="{{ site.baseurl }}/tutorials/"
-          style="font-weight: bold;"
       {% else %}
           href="{{ site.baseurl }}{{component_url}}#main-content"
       {% endif %}
       >
       <img class="item_tutorial-image-link"
-           src="{{ site.baseurl }}/public/thumbnails/{{ component }}">
+           src="{{ site.baseurl }}/public/thumbnails/{{ component }}"
+           alt="">
```
aaruni96, line 57: Is there a point to having an empty alt string ?

_includes/tutorial.html
```diff
@@ -83,7 +83,8 @@ <h3>Available Tutorials</h3>
   <div class="item_tutorial_card {{post.test_status}}">
     <div class="item_tutorial_left">
         <img class="item_tutorial-image"
-             src="{{ site.baseurl }}/public/thumbnails/{{ post.thumbnail }}">
+             src="{{ site.baseurl }}/public/thumbnails/{{ post.thumbnail }}"
+             alt="">
```
aaruni96, line 88: Same as above

_includes/tutorial.html
```diff
@@ -110,7 +111,8 @@ <h3>Available Tutorials</h3>
   <div class="item_tutorial_card {{post.test_status}}">
     <div class="item_tutorial_left">
         <img class="item_tutorial-image"
-             src="{{ site.baseurl }}/public/thumbnails/{{ post.thumbnail }}">
+             src="{{ site.baseurl }}/public/thumbnails/{{ post.thumbnail }}"
```
aaruni96, line 115: Same as above

events/meetings.md
```diff
@@ -15,7 +15,9 @@ meeting: true
       <em>{{ event.start-date | date: site.full_date_format }} – {{ event.end-date | date: site.full_date_format }}</em><br>
       Location: {{ event.location }}<br>
       {% if event.website %}
-        <a href="{{ event.website | replace: "https://www.oscar-system.org", site.baseurl }}" target="_blank">More information</a>
+        {% comment %}Use baseurl for OSCAR-hosted pages so links also work in local and GitHub Pages builds; open only external sites in a new tab.{% endcomment %}
```
aaruni96, line 18: Why not just use HTML style comments ( `<!-- comment -->` ) like you've used further down in this PR ?

_includes/tutorial.html
HereAround: My understanding is that the original code produces two style attributes when the tutorial is selected. That is invalid HTML, and the browser may ignore the bold styling that is being invoked a few lines below. To avoid this, the proposed change combines the two into a single style attribute.

_includes/sidebar.html
HereAround: The link opens a page in a new tab.
* noopener prevents the newly opened page from controlling the OSCAR page.
* noreferrer prevents the newly opened page from seeing which page the visitor came from.

See for instance https://www.timospecht.de/en/what-is/rel-noopener-noreferrer/ (but I am sure there are other and potentially better descriptions).

_includes/tutorial.html
HereAround: My understanding is: yes.

For context, this concerns screen readers, which read webpage content aloud for visually impaired users.

`alt=""` tells screen readers to ignore the image because it is decorative. The picture is decorative, because the tutorial’s topic is already written next to the image. Hence, reading out an additional description would be redundant.

_includes/tutorial.html
HereAround: Same as above.

_includes/tutorial.html
HereAround: Same as above.

events/meetings.md
HereAround: Two points:
1. Your observation is correct - the changes were inconsistent.
2. I believe that none of the added comments need to appear on the website, nor its underlying, generated HTML source.
 
With the resolve for 2), I have now changed all comments to employ Liquid comment syntax ({% comment %}...); unless I am mistaken, Jekyll removes such comments when building the website and so these comments will not appear in the generated HTML code (and consequently also not on the website).

No strong opinion as long as these comments do not appear on the actual website. As such, if you prefer, I am happy to use HTML style comments.

aaruni96: Personally, I prefer HTML style comments. Subjectively to my eyes, they look "nicer". But this is not super important, as long as things stay consistent.

_includes/tutorial.html
aaruni96: Then perhaps something like this should be the solution to not have the source be "ugly" ?

```diff
diff --git a/_includes/tutorial.html b/_includes/tutorial.html
index 1543a40..e0b7987 100644
--- a/_includes/tutorial.html
+++ b/_includes/tutorial.html
@@ -44,8 +44,9 @@ If you have not yet installed OSCAR, follow our <a href="{{site.baseurl}}/instal
   {% assign topic_tutorial = site.data.tutorials | where: "thumbnail", component | first %}
   {% assign component_url = "/tutorials/" | append: component_name | append: "/" %}
   <div class="tutorial">
-    <a style="display: block; text-align: center;{% if page.url == component_url %} font-weight: bold;{% endif %}"
+    <a
       {% if page.url == component_url %}
+          class="active"
           href="{{ site.baseurl }}/tutorials/"
       {% else %}
           href="{{ site.baseurl }}{{component_url}}#main-content"
diff --git a/_sass/oscar.scss b/_sass/oscar.scss
index 15fc039..a756036 100644
--- a/_sass/oscar.scss
+++ b/_sass/oscar.scss
@@ -137,6 +137,15 @@ details.spoiler [open] summary::after {
   font-size:2em;
 }
 
+.tutorial > a{
+  display: block;
+  text-align: center;
+}
+
+.tutorial > a.active {
+  font-weight: bold;
+}
+
 .borked{
   background: #e4e4e7;
 }
```

HereAround: Maybe like this? https://github.com/oscar-system/oscar-website/pull/885/commits/6dfd424c249c77ed66e696aa0482e9ac925ea0fd

aaruni96 (approved):
LGTM, thanks!

PR #886, opened 2026-08-02
Harden webhook and metadata update automation

HereAround:
Require a configured webhook secret, verify every request, and accept content types with parameters. Update release metadata atomically, preserve tutorial dates on API failures, normalize test results, and correct the contributor workflow filename.


etc/update_tutorials.py
```diff
@@ -38,11 +38,11 @@
         repo = g.get_repo(reponame)
         commits = repo.get_commits(path=filepath, sha=branch)
         commit = commits[0]
-        dt = commit.stats.last_modified_datetime
+        dt = commit.commit.author.date
     except Exception as e:
         print(e)
-        print("Using fallback default date for the tutorial.")
```
aaruni96, line 45: This change makes a problem less visible. We don't read logs regularly, but only when there is a problem. If we spot on the tutorial website that a tutorial was last modified in 1970, we immediately see a problem. However, if we keep "whatever was the last known date", then except if someone is really checking to see that their favorite tutorial was successfully updated, it might go unnoticed for much longer.

etc/update_tutorials.py
```diff
@@ -38,11 +38,11 @@
         repo = g.get_repo(reponame)
         commits = repo.get_commits(path=filepath, sha=branch)
         commit = commits[0]
-        dt = commit.stats.last_modified_datetime
```
aaruni96, line 42: Practically speaking, what is the difference between last-modified-datetime and author.date ?

webhook.php
```diff
@@ -39,10 +39,13 @@
...
     die();
 });
+if ($hookSecret === false || $hookSecret === '') {
+    throw new \Exception('GITHUB_WEBHOOK_SECRET is not configured.');
+}
 
 $rawPost = NULL;
 
-if ($hookSecret !== NULL) {
+{
```
aaruni96: Earlier you check for false or empty-string. I am not convinced that is a good replacement for checking for "not NULL".

webhook.php
```diff
@@ -39,10 +39,13 @@
     echo "Error on line {$e->getLine()}: " . htmlSpecialChars($e->getMessage());
     die();
 });
+if ($hookSecret === false || $hookSecret === '') {
```
aaruni96: In principle we can add this, yes. On the other hand, do we actually need this? Documentation says how a secret must be configured, and this is not exactly a public facing utility......

aaruni96 (changes requested):

etc/update_tutorials.py
HereAround: You are right. I am restoring the 1970 fallback.

I think the original 1970 fallback was broken; `datetime` must be imported, no? I will add this.

etc/update_tutorials.py
HereAround: My understanding is this:
- `author.date`: “I wrote this change on Thursday.”
- `last_modified_datetime`: “This change was added to the repository on Friday.”

On closer thought, I think we want to use `last_modified_datetime`. This very same information we can access with `committer.date` via the Git commit metadata. This would be an improvement, as I see it; I shall modify accordingly momentarily.

webhook.php
HereAround: Better safe than sorry?

webhook.php
HereAround: My understanding is this:

* `getenv()` returns false when the variable is not configured; it does not return NULL.
* `getenv()` can return an empty string when the variable is configured without a value.
 
The current `!== NULL` check will not detect either case.

etc/update_tutorials.py
```diff
@@ -67,6 +69,7 @@
     status = job.conclusion
     if status is None:
         continue
+    status = "success" if status == "success" else "failure"
```
aaruni96: If I remember the API correctly, a job's conclusion can be "success" or it can be "failure". Its possible it could also be "cancelled" or "skipped", but I will have to re-read the API docs to check that.

Anyway, in our use case, we want to check if the tutorial actually failed testing. If the testing didn't succeed for some other reason (so job conclusion is something other than "failure") we should err on the side of caution and not mark it as failed. As a reminder, any tutorials which this script marks as failed are hidden by default in the website. (This was the logic at some point, but checking right now, we changed it to explicitly checking for "success").

We should revert this line, and also, in the tutorials layout page, we should check for "not failure" when deciding to show tutorials.

aaruni96 (commented):
I don't know why github held back this comment as "pending" and did not push this out with the rest of my review.

etc/update_tutorials.py
HereAround: Like this: https://github.com/oscar-system/oscar-website/pull/886/commits/178d2831d6197338ae9bd8c2151965f061ac9701?

etc/fetch_oscar_release_information.py
```diff
@@ -17,13 +17,6 @@
 DATAPATH = os.path.join(REPOPATH, "_data")
 RELEASEFILEPATH = os.path.join(DATAPATH, "release.yml")
 
-# Delete the release file should it exist
-try:
```
aaruni96, line 21: Why do we refactor this ?

HereAround: The purpose is to preserve the last known-good release file if updating it fails.

Currently, the script deleted the existing file before fetching the new release information. If fetching or processing then failed (for example because of a network outage) the script exited without leaving a release file.

With this change, the existing file remains in place while the new information is fetched and written to a temporary file. Only after those steps succeed do we replace the existing release file.

aaruni96: This try-except block was a rough way of internal sanity checking: `RELEASEFILEPATH` must always exist, and if for whatever reason it didn't exist when the script ran, it complained in logs. We can engineer better solutions, but I would not be in favor of dropping existing checks.

If you really want to get rid of `os.remove()`, then we should add in something which really just checks that the file exists already.

etc/update_tutorials.py
aaruni96: Assuming that works (my liquid isn't very fluent), yes that works.

etc/fetch_oscar_release_information.py
HereAround: No strong feelings here; we can revert this change if you feel the current setup is better.

To avoid misunderstandings; could you make a suggestion on how you would change this code (if at all)?

aaruni96: Something like 

```python
    if not os.path.exists(RELEASEFILEPATH):
        print(f"{RELEASEFILEPATH} was not found! This may indicate a weird bug of some sort. The script will still continue, but you should double check that everything works as expected.")
    temporary_path=.... # the rest of the changes from line 80
```

HereAround: @aaruni96 Like this?

https://github.com/oscar-system/oscar-website/pull/886/commits/52c32f862a09de6c75fe67f1c76f14473f2e74ad

aaruni96: That looks okay. Thanks!

etc/update_tutorials.py
aaruni96: Its close enough to the last modified date, lets say.

aaruni96 (approved):
LGTM, thanks!

PR #889, opened 2026-08-04
Fix tutorial status updater

aaruni96:
The tutorial status tester has been broken for ~5 months and marking everything as "success" even when it fails. This patch fixes it.


HereAround (approved):
Very good catch! Thank you @aaruni96 !

PR #898, opened 2026-08-14
Redraw and revise organigram

fingolfin:
This graphic is still used in many talks. It is more technical and simply on a different level than OSCAR-overview.svg which we shall keep in parallel.

Compared to the original version, the new one has less garish colors and is written in LaTeX + TikZ (so it is much easier to update and tweak). It also drops the AlgebraicSolving.jl -> Singular.jl arrow, which asserted a dependency that does not exist, and now lists GAP packages and polymake applications. Plus a few other tweaks and corrections.

This is what it looks like:
OSCAR-organigram

For comparison, that's the old version:
OSCAR-Organigramm


fingolfin:
The AI usage here must be declared

Also, we should unify how this and `OSCAR-overview.tex` are "compiled", and perhaps move the sources to be next to each other.

fingolfin:
Maybe we should have a section on the website which shows these diagrams plus where to get the SVG/PDF/TEX versions so people can use them; that page can also provide links to the logos, and other such material...

fingolfin:
@benlorenz good points. I've updated it a bit.

This now also unifies both diagrams a bit (colors match a bit better, sources and generated files in the same place, uniform build setup). 

Also, as @HereAround suggested, clicking on the diagram on the About page links to it in "full size".

fingolfin:
For reference, here's also new vs. old "overview diagram". Note that I've also fixed the RNG seed so that it doesn't look different each time it gets rendered

new:
OSCAR-overview

old:
OSCAR-overview

public/figures/README.md
```diff
@@ -0,0 +1,51 @@
...
+|-------------------------|-------------------|---------------------------|
+| `OSCAR-overview.tex`    | `.pdf` and `.svg` | `about.md`                |
+| `OSCAR-organigram.tex`  | `.pdf` and `.svg` | talks; more technical     |
+
+The two are deliberately different in kind: the overview is the friendly
+picture, the organigram is the detailed one. Both stay.
+
+    make            # everything committed here
+    make clean      # LaTeX by-products
+    make OSCAR-organigram.png   # 300 dpi raster of any figure
```
HereAround, line 15:
```suggestion
    make OSCAR-organigram.svg   # build the organigram SVG and its PDF prerequisite
    make OSCAR-overview.svg         # build the overview SVG and its PDF prerequisite
    make OSCAR-organigram.png   # optional 300 dpi raster
    make OSCAR-overview.png         # optional 300 dpi raster
```

public/figures/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+Needs LuaLaTeX with TikZ and Fira Sans (TeX Live's `fira`), plus poppler
+(`pdftocairo`, `pdftoppm`).
+
+Both figures build byte for byte reproducibly, so a rebuild only shows up
+in `git diff` when the drawing really changed.
```
HereAround, line 21: Nitpick - build-environment dependency.

```suggestion
The figure builds are designed to be reproducible within a consistent build
environment, so a rebuild should only show up in `git diff` when the drawing
changed. Generated files may vary between build environments.
```

about.md
```diff
@@ -28,22 +28,11 @@ developers, and students to construct, manipulate, and compute with
 sophisticated algebraic and geometric structures in a modular, extensible, and
 expressive environment.
 
-<img src="{{ site.baseurl }}/public/OSCAR-overview.png" alt="OSCAR Overview" width="70%" style="margin-left:13%;" align="center">
+<a href="{{ site.baseurl }}/public/figures/OSCAR-overview.svg" title="Show full size">
+<img src="{{ site.baseurl }}/public/figures/OSCAR-overview.svg" alt="OSCAR Overview" width="70%" style="margin-left:13%;" align="center">
+</a>
```
HereAround, line 33: Potential performance regression: the current PNG is roughly 300 KiB, while the proposed SVG is about 1,080 KiB, i.e. approximately 3.7 times larger. Thus, the About page is expected to load more slowly.

I recall that @aaruni96 and I optimized the current PNG for a good balance between quality and file size. @aaruni96, what do you think? Is the new file size acceptable, or would you recommend further optimization?

fingolfin: SVGs can be transferred gzipped (or if the browser is new enough, brotli compressed), so in practice this shouldn't actually matter, but of course it should be measured -- specifically, what  is *really* transferred over the wire.

aaruni96: I have verified that we do gzip things before sending transmitting them. I have also verified that when github sends the svg, it gzips the 1.11 MB file to 164.78 KB. Its safe to assume our compression system will do the same (or something very close). In short, this should be good enough.

HereAround (dismissed):
LGTM and is IMHO ready to be merged. (Left a few comments with nitpicks.)

HereAround:
Ready to merge IMHO (modulo nitpicks above); adding triage label for final check (hear opinions/raise awareness) before hitting merge.

HereAround (approved):

HereAround:
Triage touched on this PR. People were informed about the change in this PR. Nobody objected. I am thus merging this PR now.

